PR #56, opened 2020-08-30
add fixtures and tests for group management views

jeriox:


jeriox:
replaced by #156 and #168

PR #99, opened 2020-09-19
add QualificationGrantFormset to UserProfileUpdateView

### LINKED ISSUE #75: Manage qualification grants for user
As a Manager I want to manage qualification grants while creating or editing a user. I expect the page to have formset with forms for qualification grants. The formset is pre-populated with the existing grants for that user (if editing a user) with the option to delete each grant. I also want to have a button to enter a new grant. I expect that this adds a new form to the formset where I can enter the desired qualification and the expiration date.

Hint: have a look at pretix, they use a jquery library to manage the client-side stuff of the formsets (adding a new form, deleting)

jeriox:
closes #75


static/jep/js/main.js
```diff
@@ -1,4 +1,8 @@
 
 $(function () {
   $('[data-toggle="tooltip"]').tooltip()
-})
\ No newline at end of file
+})
+
+$('#formset').on('formAdded', function(event) {
```
jeriox, line 6: @felixrindt this event listener is never executed, but the code inside would do the right thing. could you please have a look?

ephios/static/ephios/js/main.js
```diff
@@ -1,4 +1,10 @@
...
-})
\ No newline at end of file
+    $('[data-toggle="tooltip"]').tooltip();
+
+    $("[data-formset]").formset({
+        animateForms: true,
+        reorderMode: 'dom',
+    }).on("formAdded", "div", function (event) {
+        $(event.target).find("select").first().djangoSelect2()
+    });
```
felixrindt, line 9: So this fixes your JS and also avoids inline-JS.
Although you still should change the `select` selector to something more specific for this to be in `main.js`.

ephios/user_management/templates/user_management/userprofile_form.html
```diff
@@ -21,10 +21,64 @@ <h1>{% trans "Add new user" %}</h1>
...
+                <!-- New forms will be inserted in here -->
+                {% for form in qualification_formset %}
+                    <div class="card mb-3" data-formset-form>
+                        <div class="card-body clearfix">
+                            {{ form.id }}
+                            <button class="btn btn-link float-right pt-4" type="button" data-formset-delete-button><span
+                                class="fas fa-trash-alt"></span> {% trans "Delete" %}</button>
+                            <div class="form-row">
+                                <div class="col">
+                                    {% bootstrap_field form.qualification %}
```
felixrindt, line 40: I think it makes no sense for an existing grant to be modified. Changing the qualification would better be a *new* grant. Changing the expires field might be required, but you could create a new grant still, then. Deletion is required on the other hand.

ephios/user_management/templates/user_management/userprofile_form.html
```diff
@@ -21,10 +21,64 @@ <h1>{% trans "Add new user" %}</h1>
     </div>
     <form method="post" class="form">
         {% csrf_token %}
-        {% bootstrap_form form %}
+        {% bootstrap_form userprofile_form %}
+        {% load formset_tags %}
+
+        <div class="formset" data-formset data-formset-prefix="{{ qualification_formset.prefix }}">
```
felixrindt, line 27: Can we make all qualifciations one single card and let them be list-group-items? I think that would look more apropriate, as it would section the user form better as a whole.

ephios/user_management/templates/user_management/userprofile_form.html
```diff
@@ -21,10 +21,64 @@ <h1>{% trans "Add new user" %}</h1>
...
+                            </div>
+                            <div class="d-none">{{ form.DELETE }}</div>
+                        </div>
+                    </div>
+                {% endfor %}
+            </div>
+
+            <!-- The empty form template. By wrapping this in a <script> tag, the
+            __prefix__ placeholder can easily be replaced in both attributes and
+            any scripts -->
```
felixrindt, line 54: Remove the comments for merging.

ephios/user_management/templates/user_management/userprofile_form.html
felixrindt: also the add-button could be smaller then :joy:

ephios/user_management/views.py
```diff
@@ -21,47 +30,75 @@ class UserProfileListView(PermissionRequiredMixin, ListView):
...
+        kwargs.setdefault("userprofile_form", UserProfileForm(self.request.POST or None))
+        kwargs.setdefault(
+            "qualification_formset", QualificationGrantFormset(self.request.POST or None)
+        )
+        return super().get_context_data(**kwargs)
+
+    def post(self, request, *args, **kwargs):
+        userprofile_form = UserProfileForm(self.request.POST)
+        qualification_formset = QualificationGrantFormset(self.request.POST)
+        if userprofile_form.is_valid() and qualification_formset.is_valid():
```
felixrindt, line 48: I'm not sure we'd get all validation-errors with the `and` short-circuiting. Could do: 
```suggestion
        if all((userprofile_form.is_valid(), qualification_formset.is_valid())):
```

ephios/user_management/views.py
```diff
@@ -21,47 +30,75 @@ class UserProfileListView(PermissionRequiredMixin, ListView):
...
+        self.object = self.get_object()
+        kwargs.setdefault("userprofile_form", self.get_userprofile_form())
+        kwargs.setdefault("qualification_formset", self.get_qualification_formset())
+        return super().get_context_data(**kwargs)
+
+    def post(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        userprofile_form = self.get_userprofile_form()
+        qualification_formset = self.get_qualification_formset()
+        if userprofile_form.is_valid() and qualification_formset.is_valid():
```
felixrindt, line 89: same short-circuit problem as above.

felixrindt (changes requested):

static/jep/js/main.js
felixrindt: /me gets their magic javascript gloves

ephios/static/ephios/js/main.js
jeriox: but why? enabling select2 features on newly added select inputs seems like a good idea to me

felixrindt: but I might not want *all* select input elements to be a select2 one. I'd rather have that stated explicitly, as it might lead to bugs or unwanted overhead later.

ephios/user_management/templates/user_management/userprofile_form.html
```diff
@@ -21,10 +21,67 @@ <h1>{% trans "Add new user" %}</h1>
...
+                        <!-- New forms will be inserted in here -->
+                        {% for form in qualification_formset %}
+                            <div class="list-group-item clearfix" data-formset-form>
+                                {{ form.id }}
+                                <button class="btn btn-link float-right pt-4" type="button"
+                                        data-formset-delete-button><span
+                                        class="fas fa-trash-alt"></span> {% trans "Delete" %}</button>
+                                <div class="form-row">
+                                    <div class="col">
+                                        {% bootstrap_field form.qualification %}
```
felixrindt, line 43: have you tried this?
```suggestion
                                        {% bootstrap_field form.qualification layout="horizontal" %}
```

felixrindt (commented):
I like the looks more now, thanks! :+1:

felixrindt:
You could add a box to the user's profile page with a list of grants while you're at it... :)

ephios/user_management/templates/user_management/userprofile_form.html
felixrindt: (just so it takes up less vertical space)

ephios/static/ephios/js/main.js
felixrindt: The django-select2 widgets already come with a `django-select2` class, so I'd suggest to use `.django-select2` as the jquery selector.

felixrindt: `$('.django-select2').djangoSelect2()` is also what the packaged JS does.

ephios/user_management/templates/user_management/userprofile_form.html
jeriox: good idea, makes it even smaller

jeriox:
> You could add a box to the user's profile page with a list of grants while you're at it... :)

no, won't do. this is definetly out of scope for this story and is already planned for the next release ;)

ephios/static/ephios/js/main.js
```diff
@@ -1,4 +1,10 @@
...
-  $('[data-toggle="tooltip"]').tooltip()
-})
\ No newline at end of file
+    $('[data-toggle="tooltip"]').tooltip();
+
+    $("[data-formset]").formset({
+        animateForms: true,
+        reorderMode: 'dom',
+    }).on("formAdded", "div", function (event) {
+        $(event.target).find(".django-select2").first().djangoSelect2()
```
felixrindt, line 8: You don't meed the `first`

felixrindt (approved):
No with the select being disabled, a user might think there *is* a way to enable them, but there is not. What do you think about just putting a text there with the qualification?

felixrindt:
> no, won't do. this is definetly out of scope for this story and is already planned for the next release ;)

Ok, I don't see a Github issue for that. An also, why not put it in this release? Seems pretty essential to me that users know what qualifications they have.

jeriox:
> No with the select being disabled, a user might think there *is* a way to enable them, but there is not. What do you think about just putting a text there with the qualification?

I thought about that, there is even a bootstrap class for it. But django-bootstrap offers no way of setting classes on the field itself. This is not the first time that it limits our ideas. Maybe we should switch to crispy-forms or smth like that

jeriox:
> > no, won't do. this is definetly out of scope for this story and is already planned for the next release ;)
> 
> Ok, I don't see a Github issue for that. An also, why not put it in this release? Seems pretty essential to me that users know what qualifications they have. 

Yes, I didn't create issues for the next release yet.
I don't think so. Everyone got around without knowing their qualifications during the last years. And the first release was intended to deliver the old functionality..

felixrindt:
> > No with the select being disabled, a user might think there _is_ a way to enable them, but there is not. What do you think about just putting a text there with the qualification?
> 
> I thought about that, there is even a bootstrap class for it. But django-bootstrap offers no way of setting classes on the field itself. This is not the first time that it limits our ideas. Maybe we should switch to crispy-forms or smth like that

Can you be more explicit about the problem you are trying to solve? There might be a way you don't know of ;)

felixrindt:
> > > no, won't do. this is definetly out of scope for this story and is already planned for the next release ;)
> > 
> > 
> > Ok, I don't see a Github issue for that. An also, why not put it in this release? Seems pretty essential to me that users know what qualifications they have.
> 
> Yes, I didn't create issues for the next release yet.
> I don't think so. Everyone got around without knowing their qualifications during the last years. And the first release was intended to deliver the old functionality..

yea fine I'm just going to do it.

PR #108, opened 2020-09-22
show qualifications on user profile detail

felixrindt:
image


felixrindt:
I'm not sure about the plural form here. Might as well be "Qualification" just like "information".

jeriox (approved):

PR #109, opened 2020-09-22
Pages plugin

### LINKED ISSUE #102: Footer information
As an internet user, I expect there to be a footer containing links to Legal information, data protection information and a link to the software used.

We could go the route of creating a Markdown-mini-crm-Pages model that can be enabled to be visible publicly and in the footer.
Or we hardcode something.

felixrindt:
Fix #102


felixrindt:
image

-------------

image

ephios/templates/base.html
```diff
@@ -102,6 +102,14 @@
             {% block content %}{% endblock %}
         </div>
     </div>
+    <footer class="footer text-center mt-auto py-3">
+        <small>
+            {% for label, url in footer.items %}
+                <a href="{{ url }}">{{ label }}</a> ·
+            {% endfor %}
+            <span class="text-muted">powered by <a href="https://ephios.de/" target="_blank">ephios</a></span>
```
jeriox, line 110: translation?

ephios/templates/base.html
```diff
@@ -102,6 +102,14 @@
             {% block content %}{% endblock %}
         </div>
     </div>
+    <footer class="footer text-center mt-auto py-3">
```
jeriox, line 105: i'd prefer to have a footer that sticks to the bottom of the viewport (like so: https://getbootstrap.com/docs/4.1/examples/sticky-footer/)

ephios/plugins/pages/views.py
```diff
@@ -0,0 +1,20 @@
+from django.views.generic import DetailView
+
+from ephios.plugins.pages.models import Page
+
+from django.core.exceptions import PermissionDenied
+
+
+class PageView(DetailView):
+    model = Page
+    accept_global_perms = True
```
jeriox, line 10: do we need this?

jeriox (commented):
works like a charm!

ephios/templates/base.html
jeriox: (but with your design, their big grey bar is ugly)

felixrindt: I'll try :)

ephios/plugins/pages/views.py
felixrindt: nope

jeriox (approved):
looks better, but the viewport stuff is still a bit buggy in chrome on android

PR #112, opened 2020-09-23
add content security policy

### LINKED ISSUE #95: Implement Content Security Policy
which means we can't have inline javascript!

* https://docs.djangoproject.com/en/3.1/topics/security/
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src

jeriox:
closes #95


ephios/event_management/templates/event_management/event_archive.html
```diff
@@ -1,27 +1,25 @@
...
-    {% include "event_management/fragments/datatable.html" with order_column=2 order_direction="desc" %}
-{% endblock %}
-
 {% block content %}
     <div class="page-header">
         <h1>{% trans "Past events" %}</h1>
     </div>
-    <table id="event_table" class="table table-striped display datatable">
+    {% get_current_language as current_language %}
+    <table id="event_table" class="table table-striped display datatable" data-order="[[ 2, &quot;desc&quot; ]]" {% if current_language == "de-de" %} data-language="{&quot;url&quot;: &quot;{% static "datatables/german.json" %}&quot;}" {% endif %} >
```
felixrindt, line 15: This line looks really hacky and is specific to German. Can you explain what's going on and why it needs to be this way?

ephios/event_management/templates/event_management/fragments/shift_box_big.html
```diff
@@ -41,11 +41,8 @@ <h5 class="card-title">
                                                      <p>{{ error.message }}</p>
                                                  {% endfor %}"
                                                 {% endif %}>
-                                            <button type="submit" class="btn btn-outline-success"
-                                                    name="signup_choice" value="sign_up"
-                                                    {% if not can_sign_up %}
-                                                    style="pointer-events: none;" disabled
-                                                    {% endif %}>
+                                            <button type="submit" name="signup_choice" value="sign_up"
+                                                    class="btn btn-outline-success {% if not can_user_decline %}btn-disabled" disabled {% else %}"{% endif %}>
```
felixrindt, line 45: should be "can_sign_up" here?

Also can we rename the other one to `can_decline` (remove `user`)

felixrindt: (Rename includes the template tags right?)

ephios/static/ephios/js/main.js
```diff
@@ -1,4 +1,4 @@
-$(function () {
+$(document).ready(function () {
```
felixrindt, line 1: Is there a semantic difference to `$()`?

ephios/settings.py
```diff
@@ -65,6 +65,7 @@
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "csp.middleware.CSPMiddleware",
```
felixrindt, line 68: I'm gonna read some of the documentation for django-csp. 
Is the default configuration everything we want?

felixrindt (changes requested):

ephios/event_management/templates/event_management/event_archive.html
jeriox: you need to tell datatables which language to use. Unfortunately they dont't stick to the short codes for languages that django uses. I just moved it here from the old template snippet and thought that it's gonna be okay for the moment as we only have german translations anyway.

ephios/static/ephios/js/main.js
jeriox: TIL no. I prefer the verbose variant anyway because this is what tutorials use. never heard of the short variant to date.

ephios/settings.py
jeriox: yes. this restricts everything to self.

felixrindt: nice

ephios/event_management/templates/event_management/fragments/shift_box_big.html
```diff
@@ -41,28 +41,22 @@ <h5 class="card-title">
                                                      <p>{{ error.message }}</p>
                                                  {% endfor %}"
                                                 {% endif %}>
-                                            <button type="submit" class="btn btn-outline-success"
-                                                    name="signup_choice" value="sign_up"
-                                                    {% if not can_sign_up %}
-                                                    style="pointer-events: none;" disabled
-                                                    {% endif %}>
+                                            <button type="submit" name="signup_choice" value="sign_up"
+                                                    {% if not can_sign_up %}class="btn btn-outline-success btn-disabled" disabled>{% else %}class="btn btn-outline-success btn-disabled">{% endif %}
```
felixrindt, line 45: Please review this line. You have `btn-disabled` in both cases. Also, lets remove the `not` to avoid confusion. (same for decline)

ephios/event_management/templates/event_management/event_archive.html
felixrindt: and we can't do that in our databales initializer in JS?

jeriox: could do, but then we'd need to pass the current_language from django to JS which is a bit ugly with the CSP enabled

felixrindt: It's needn't be ugly, I've seen that before. Put a `data-language` tag on the html `body` element and read that from JS. :)

felixrindt (approved):

PR #122, opened 2020-09-23
fix admin email setting

felixrindt:
Test with `./manage.py sendtestmail --admin`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,4 +9,4 @@ SITE_URL=http://localhost:8000
 EMAIL_URL=dummymail://
 DEFAULT_FROM_EMAIL=webmaster@localhost
 SERVER_EMAIL=root@localhost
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -9,4 +9,4 @@ SITE_URL=http://localhost:8000
 EMAIL_URL=dummymail://
 DEFAULT_FROM_EMAIL=webmaster@localhost
 SERVER_EMAIL=root@localhost
-ADMINS=("root", "root@localhost")
+ADMINS=Root <root@localhost">
```
jeriox, line 12: only one " looks wrong

felixrindt: WHY DOES IT WORK THEN I ASK YOU (fixed)

jeriox (approved):

PR #126, opened 2020-09-24
Fixes and just like nice things

### LINKED ISSUE #127: Poetry improvements
* [ ] https://python-poetry.org/docs/pyproject/#extras for mysql, postgres database backends
* [x] show version in footer
* [ ] ~~define plugin entry point~~ Poetry can do that, but that will happen no earlier than v0.2.0

felixrindt:
A wild mix of small stuff:

* fix #127 
* rename extra to helpers
* fix password reset
* show version in footer
* add dependency extra groups for mysql and pgsql
* isort


ephios/extra/context.py
```diff
@@ -1,7 +1,16 @@
...
 from django.utils.translation import get_language
 
-from ephios.helpers.signals import footer_link
+from ephios.extra.signals import footer_link
+
+import subprocess
+
+try:
+    EPHIOS_VERSION = (
+        subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode().strip()
```
jeriox, line 10: Can we cache that so we don't need to wait for the subproccess on every page load?

felixrindt: It's already only done at module load?!

ephios/settings.py
```diff
@@ -107,7 +106,7 @@
         "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator",
     },
     {
-        "NAME": "user_management.utils.CustomMinimumLengthValidator",
+        "NAME": "ephios.user_management.utils.CustomMinimumLengthValidator",
```
felixrindt, line 109: That's what broke the password reset in the call ;)

ephios/extra/context.py
```diff
@@ -0,0 +1,32 @@
...
+
+from ephios.extra.signals import footer_link
+
+try:
+    EPHIOS_VERSION = (
+        subprocess.check_output(["git", "rev-parse", "--short", "HEAD"]).decode().strip()
+    )
+except (subprocess.CalledProcessError, FileNotFoundError):
+    # suggested in https://github.com/python-poetry/poetry/issues/273
+    EPHIOS_VERSION = f"v{importlib.metadata.version('ephios')}"
```
felixrindt, line 15:
```suggestion
    EPHIOS_VERSION = "v" + importlib.metadata.version('ephios')
```

jeriox (approved):

PR #128, opened 2020-09-24
Event times getter

felixrindt:
Ok hear me out on this one. This is about performance. When we have 100 events in the list, we don't want to make queries for every single one of them. But that's what has been happening with start and end times. I now changed start and end time on event to be a getter to make more clear that it does a query. (It was only used in the multi shift pdf export afais).

Then I just used annotate for the two list views.


felixrindt:
Btw you can test for those issues by having N elements in a list and checking that there are less than N queries performed (with N sufficiently high, e.g. 50)

jeriox (approved):

PR #135, opened 2020-09-25
add event type to event lists

jeriox:


felixrindt (approved):
add a select related for event type in both `get_queryset`s

PR #141, opened 2020-09-25
Fixes

### LINKED ISSUE #129: Multiple mails for participation
When requesting a participation for a shift for the first time, two identical mails are sent to the user stating that the participation is now requested. This is probably due to the two steps of creating the participation and setting the state of the participation (which already has the default state of requested), both times triggering the post-save signal

felixrindt:
* Event form permissions
* "Requests" translation
* Editing the user with existing qualification
* fix #129

### DIFF
--- a/ephios/event_management/signup.py
+++ b/ephios/event_management/signup.py
@@ -86,7 +86,7 @@ class LocalUserParticipant(AbstractParticipant):
     user: get_user_model()
 
     def create_participation(self, shift):
-        return LocalParticipation.objects.create(shift=shift, user=self.user)
+        return LocalParticipation(shift=shift, user=self.user)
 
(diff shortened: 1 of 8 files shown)


ephios/event_management/signup.py
```diff
@@ -86,7 +86,7 @@ class LocalUserParticipant(AbstractParticipant):
     user: get_user_model()
 
     def create_participation(self, shift):
-        return LocalParticipation.objects.create(shift=shift, user=self.user)
+        return LocalParticipation(shift=shift, user=self.user)
```
jeriox, line 89: maybe rename the method? could be misleading because now the participation is not persisted after creation

jeriox (approved):

PR #145, opened 2020-09-26
Fix #144: Custom Permssion Required Mixin

### LINKED ISSUE #144: Login loop instead of PermissionDenied
When accessing a page as a logged in user for which I don't have the appropriate permission, I am redirected to the login page. After a successful login, this happends again in an endless loop. I expect the PermissionRequiredMixin to behave in the same way as the django PermissionRequiredMixin which states 

```        
if self.raise_exception or self.request.user.is_authenticated:
            raise PermissionDenied(self.get_permission_denied_message())
return redirect_to_login(self.request.get_full_path(), self.get_login_url(), self.get_redirect_field_name())
```

felixrindt:
Fixes #144

Implemented a custom mixin which also defaults to accept global perms.
I really liked the django mixin implementation, so I decided to base it on that. 
It actually had a designated method to overwrite :)


felixrindt:
I should remove more of the `accept_global-perms` attributes

ephios/user_management/migrations/0003_userprofile_calendar_token_squashed_0008_auto_20200925_1640.py
```diff
@@ -1,11 +1,12 @@
 # Generated by Django 3.1.1 on 2020-09-25 14:54
 
-from django.conf import settings
-from django.db import migrations, models
-import django.db.models.deletion
 import secrets
 import uuid
 
+import django.db.models.deletion
```
jeriox, line 6: do you apply isort locally? maybe you should add it to the repo ^^

jeriox (approved):

ephios/user_management/migrations/0003_userprofile_calendar_token_squashed_0008_auto_20200925_1640.py
felixrindt: yes I have it on my list :)

PR #156, opened 2020-09-27
Fix #148, with tests and other fixes

### LINKED ISSUE #148: Responsible groups are displayed in visible_for
When editing an event, responsible groups are displayed in the visible_for field although they were never added to visible_for. this is due to the fact that they are assigned the view_event permission in case they aren't in any groups that the event is visible for. We should exclude groups that are responsible from the initial set for the visible_for field

felixrindt:
Fix #148


ephios/event_management/forms.py
```diff
@@ -46,6 +47,41 @@ class Meta:
         model = Event
         fields = ["title", "description", "location", "type", "mail_updates"]
 
+    def __init__(self, **kwargs):
+        user = kwargs.pop("user")
+        can_publish_for_groups = get_objects_for_user(user, "publish_event_for_group", klass=Group)
+
+        if (event := kwargs.get("instance", None)) is not None:
+            responsible_persons = get_users_with_perms(
```
jeriox, line 55: should rename to responsible_users here as well

ephios/event_management/views.py
```diff
@@ -63,51 +63,28 @@ class EventUpdateView(CustomPermissionRequiredMixin, UpdateView):
...
     def get_form(self, form_class=None):
-        visible_for_queryset = get_objects_for_user(
-            self.request.user, "publish_event_for_group", klass=Group
-        )
-        event_form = EventForm(
-            self.request.POST or None,
+        return EventForm(
+            data=self.request.POST or None,
+            user=self.request.user,
             initial={
```
jeriox, line 80: this is uneccessary as the form handles this now AFAIS?

ephios/event_management/forms.py
```diff
@@ -66,12 +102,13 @@ def save(self, commit=True):
             Group.objects.filter(
                 Q(id__in=self.cleaned_data["visible_for"])
                 | Q(id__in=self.cleaned_data["responsible_groups"])
+                | Q(id__in=self.locked_visible_for_groups)
```
jeriox, line 105: this is broken. steps to reproduce:
- have an event that is visible for volunteers and managers and planners are responsible
- visit the edit view with a user in the planner group
- you'll see volunteers in the select field for visible and "also for manager" in the help text
- remove volunteers from the visibility and save the page
- now the event is not visible for managers anymore

jeriox (changes requested):
yay tests <3

ephios/event_management/forms.py
jeriox: the own user that is prefilled in initial still does not get saved/assigned change permissions. not sure if this is related to this renaming issue

jeriox (changes requested):

ephios/event_management/views.py
felixrindt: well no. this is the create view. the form only knows how to initialize stuff with a given instance.

ephios/event_management/forms.py
felixrindt: it probably was

ephios/event_management/forms.py
felixrindt: Very good catch. The locked groups were stored as an *unevaluated* queryset. Evaluation only happened after all permissions had been removed, so the queryset was empty. It is now initialized on form initialization.

jeriox (approved):

PR #173, opened 2020-10-14
update dependencies to be compatible with python3.9

jeriox:


pyproject.toml
```diff
@@ -9,7 +9,7 @@ homepage = "https://ephios.de"
 repository = "https://github.com/ephios-dev/ephios"
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = "^3.9"
```
felixrindt, line 12: to me, this looks like python 3.9 only. I don't think we want to drop support for 3.8?!

jeriox: yes, this is 3.9 only. I thought there were some fancy new features (or syntax?) that we wanted to use? I thought that would require dropping support for 3.8

felixrindt (approved):

PR #176, opened 2020-10-16
Manage permissions for group

### LINKED ISSUE #142: Manage permissions for group
As a manager, when editing groups I want to have checkboxes fot "can manage user" and "can manage groups" that control the appropriate permissions on the group.

gritfessel:
Closes #142


ephios/user_management/views.py
```diff
@@ -175,6 +175,22 @@ def get_form_kwargs(self):
...
+            ).exists(),
+            "can_manage_group": self.object.permissions.filter(
+                codename__in=[
+                    "add_group",
+                    "change_group",
+                    "delete_group",
+                    "view_group",
+                ]
+            ).exists(),
         }
```
felixrindt, line 194: Can you move the bulk part of this method to the form class?

felixrindt (changes requested):

ephios/user_management/forms.py
```diff
@@ -114,6 +126,36 @@ class Meta:
...
+                    ]
+                ).exists(),
+                "can_manage_group": group.permissions.filter(
+                    codename__in=[
+                        "add_group",
+                        "change_group",
+                        "delete_group",
+                        "view_group",
+                    ]
+                ).exists(),
```
felixrindt, line 155:
```suggestion
                ).exists(),
                **kwargs.get("initial", {}),
```

or smth. like that. Has the disadvantage that the `exists` queries here are run even if not used, but I did the same for EventForm and avoiding that would need more complex control flow.

felixrindt (changes requested):

felixrindt (approved):

PR #179, opened 2020-10-16
add view and form to duplicate event, add django-recurrence

### LINKED ISSUE #149: Duplicate event
As a planner, I want to have a button in the event management menu to duplicate an event.
I expect the corresponding page to have controls to duplicate the event to specific dates or for an interval (e.g. every monday).
After saving, I expect the new events to be created with the same shifts as the old event.

jeriox:
closes #149 

ToDo:

- [x] display selected dates in a neat way
- [x] rename to copy
- [x] add field for start date
- [x] make migration for event series
- [x] write tests


jeriox:
@felixrindt maybe you can have a look at the javascript event stuff to auto-update the preview. I think the solution with the button is okay for now otherwise

ephios/event_management/views.py
```diff
@@ -134,6 +141,82 @@ def get_queryset(self):
...
+    form_class = EventDuplicationForm
+
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        self.object = self.get_object()
+
+    def form_valid(self, form):
+        occurences = form.cleaned_data["recurrence"].between(
+            datetime.now() - timedelta(days=1),
+            datetime.now() + timedelta(days=365),
```
felixrindt, line 157: Does this mean I can't create events in 2022? Seems unwanted to me.

ephios/event_management/forms.py
```diff
@@ -143,3 +144,14 @@ def clean(self):
             if not cleaned_data["meeting_time"] <= cleaned_data["start_time"]:
                 raise ValidationError(_("Meeting time must not be after start time!"))
         return cleaned_data
+
+
+class EventDuplicationForm(Form):
+    start_date = DateField(
+        widget=CustomDateInput(format="%Y-%m-%d"),
+        initial=datetime.now().date(),
```
felixrindt, line 152: I think it might be better if this had no default and must be explicitly set.

ephios/event_management/views.py
```diff
@@ -134,6 +141,82 @@ def get_queryset(self):
...
+                    recurrence.between(
+                        datetime.now() - timedelta(days=1),
+                        datetime.now() + timedelta(days=365),
+                        inc=True,
+                        dtstart=datetime.combine(
+                            DateField().to_python(self.request.POST["dtstart"]), datetime.min.time()
+                        ),
+                    ),
+                    cls=CustomJSONEncoder,
+                )
```
felixrindt, line 214: You should format the dates at some point, either here or in JS:

image

ephios/event_management/views.py
```diff
@@ -134,6 +141,82 @@ def get_queryset(self):
...
+            )
```
felixrindt, line 215: Also, the "date_to" picker uses yet another format (yyyy-mm-dd). Anything we can do about that? I wanted to type the date manually but didn't figure out the format until I tried the terrible picker.

image

ephios/event_management/views.py
```diff
@@ -134,6 +141,82 @@ def get_queryset(self):
...
+                offset = shift.start_time.date() - start_date
+                # shifts ending on the next day should end on the next day to the new date
+                end_offset = shift.end_time.date() - shift.start_time.date()
+                shift.end_time = datetime.combine(
+                    date.date() + offset + end_offset, shift.end_time.time()
+                )
+                shift.meeting_time = datetime.combine(
+                    date.date() + offset, shift.meeting_time.time()
+                )
+                shift.start_time = datetime.combine(date.date() + offset, shift.start_time.time())
```
felixrindt, line 188: You are loosing time zone information here.

```
/home/felix/.virtualenvs/ephios/lib/python3.8/site-packages/django/db/models/fields/__init__.py:1367: RuntimeWarning: DateTimeField Shift.meeting_time received a naive datetime (2021-09-15 10:45:00) while time zone support is active.
  warnings.warn("DateTimeField %s received a naive datetime (%s)"
/home/felix/.virtualenvs/ephios/lib/python3.8/site-packages/django/db/models/fields/__init__.py:1367: RuntimeWarning: DateTimeField Shift.start_time received a naive datetime (2021-09-15 11:00:00) while time zone support is active.
  warnings.warn("DateTimeField %s received a naive datetime (%s)"
/home/felix/.virtualenvs/ephios/lib/python3.8/site-packages/django/db/models/fields/__init__.py:1367: RuntimeWarning: DateTimeField Shift.end_time received a naive datetime (2021-09-15 18:00:00) while time zone support is active.
  warnings.warn("DateTimeField %s received a naive datetime (%s)"
```

felixrindt (changes requested):
I think it's a really good start. 
The rrule lib has some UX issues, but I think this is a good way to go. We probably need some more information about the event being copied on the copy view (maybe use the small shift boxes again as information?)

jeriox:
> I think it's a really good start.
> The rrule lib has some UX issues, but I think this is a good way to go. We probably need some more information about the event being copied on the copy view (maybe use the small shift boxes again as information?)

I thought about that. Originally, I wanted to include a list of dd's for the shift, but I couldn't think of a nice way to display it as the shift info contains the date and I think that would be confusing on a page with a lot of dates. An alternatative I came up with was something like "Shift on Day 1 08:00-18:00", but I don't think that is intuitive. And you can only reach the copy view from the detail view, so you can have a look there

ephios/event_management/views.py
jeriox: Yes.
It was meant as a precaution to prevent users from creating events that recurr forever. As we dropped the event series, it will be very hard to delete all of these afterwards. Do you have a good alternative for that?

ephios/event_management/views.py
jeriox: Not really. Only way would be to fork and adapt the library, which we should do anyways (probably some time in the future)

ephios/event_management/views.py
felixrindt: Well I thought we'd add a feature to bulk delete events by filtering for type+date_from+date_to. That would enable people to undo that.

jeriox: Hm, kinda complicated. You'd need to select a start date but no end date in the filter and make sure that there are no legitimate events in between that you dont't want to delete

felixrindt: Well we need *some* kind of bulk delete. Until then, I'd suggest to increase the limit to two years, so you can at least create events for the whole upcoming year.

felixrindt: Even better, enforce that a recurrence end is set in the front end.

jeriox: limit is now two years, enforcing an end date is moved to #195

ephios/event_management/views.py
jeriox: moved to #195

jeriox:
@felixrindt i changed the limit to two years and moved all other concerns to #195 
Could you please review and approve?

gritfessel (dismissed):
- The date for "signup until" is not moved, but copied as it is in the original event. I am not sure about what you would expect the signup date to happen - either stay as it is as single sign up for the whole series of events or move analogously. Anyways, when an event is copied to an earlier date, the signup date has to change as well so that it is not after the event itself.
- When only a start date is selected in the copy view, the preview still says "no dates selected" and you cannot copy the event. I would actually expect that you could do that. If there's a good reason why you shouldn't, you should at least adjust the help text of the start date field (The event will be copied to this date.)
- Also, the loading time is pretty long depending on how many copies you make. That is not very user friendly as the user doesn't know what is happening and could try to reload or cancel the copying and therefore break things. On the long run, we need to think of something more user friendly.

felixrindt (approved):

ephios/event_management/views.py
```diff
@@ -134,6 +141,84 @@ def get_queryset(self):
...
+    form_class = EventDuplicationForm
+
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        self.object = self.get_object()
+
+    def form_valid(self, form):
+        occurences = form.cleaned_data["recurrence"].between(
+            datetime.now() - timedelta(days=1),
+            datetime.now() + timedelta(days=730),
```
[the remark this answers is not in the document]
jeriox: this is obsolete as soon as we have #195

jeriox:
@gritfessel 

- signup date moved to #209 
- start date fixed
- performance improved slighty (events cannot be bulk_created due to the foreign-key-relation)

PR #181, opened 2020-10-17
add isort to CI checks

felixrindt:


pyproject.toml
```diff
@@ -61,7 +61,6 @@ duplicate-code, attribute-defined-outside-init, missing-module-docstring, missin
 no-member, invalid-name, import-outside-toplevel, unused-argument, too-many-ancestors, missing-function-docstring,
 too-few-public-methods, no-self-use, too-many-arguments, cyclic-import, inconsistent-return-statements
 """
-confidence = "INFERENCE"
```
felixrindt, line 64: this did actually suppress a lot of helpful warnings

ephios/user_management/models.py
```diff
@@ -31,6 +31,7 @@ def create_user(
         date_of_birth,
         password=None,
     ):
+        # pylint: disable=unexpected-keyword-arg,no-value-for-parameter
         user = self.model(
             email=email,
             first_name=first_name,
```
felixrindt, line 37: Here, we are assuming `self.model` is `UserProfile`, but it doesn't has to be, which is why pylint complained I guess, just not on my local machine :/

jeriox (approved):

PR #182, opened 2020-10-17
add js/css compression

felixrindt:


felixrindt:
This needs a migration of development setup:
Move `db.sqlite3` to `data/db.sqlite3`.

@gritfessel @jeriox

ephios/settings.py
```diff
@@ -140,9 +147,15 @@
 # Static files (CSS, JavaScript, Images)
 # https://docs.djangoproject.com/en/3.0/howto/static-files/
 
-STATIC_URL = env.str("STATIC_URL")
-STATIC_ROOT = env.str("STATIC_ROOT")
+STATIC_URL = env.str("STATIC_URL", default="/static/")
+STATIC_ROOT = env.str("STATIC_ROOT", default=os.path.join(DATA_DIR, "static"))
```
jeriox, line 151: I dont't think that defaulting to data/static is what we want here. The path to the static files should be actively set when installing IMO

.env.example
```diff
@@ -1,9 +1,7 @@
 SECRET_KEY=b0jbk58)l&^hy5cht+j$n9!laoke5ivdc+3&@qf2yrt1udvb85
 DEBUG=True
-DATABASE_URL=sqlite:///db.sqlite3
+DATABASE_URL=sqlite:///data/db.sqlite3
```
jeriox, line 3: what is the purpose of this?

felixrindt: With compression enabled, static files get collected to STATIC_ROOT, which is why we need a sensible default for the development environment. If we later introduce media files, these will also need to be saved somewhere in the development setup. I propose to put these two and the database in a `data` folder by default.

When doing a proper installation, people can set a different DATA_DIR, or even customize static and media roots for serving via a webserver.

felixrindt: DATA_DIR exists as a shortcut and as a way to create the directory for initial migration. sqlite failed on me when `data` did not exist to save `db.sqlite3` there. 

To answer your other comment, we could remove the default for DATA_DIR, but if you set custom DB, static and media files location, you don't need the DATA_DIR.

ephios/templates/base.html
```diff
@@ -1,22 +1,36 @@
...
+        <link type="text/css" media="screen" rel="stylesheet" href="{% static "select2/css/select2.min.css" %}">
+        {% block css %}{% endblock %}
+    {% endcompress %}
+    {% compress js %}
+        <script type="text/javascript" src="{% static "js/jquery.min.js" %}"></script>
+        <script type="text/javascript" src="{% static "bootstrap/js/bootstrap.bundle.min.js" %}"></script>
+        <script type="text/javascript" src="{% static "select2/js/select2.js" %}"></script>
+        <script type="text/javascript" src="{% static "django_select2/django_select2.js" %}"></script>
+        <script type="text/javascript" src="{% static "datatables/datatables.min.js" %}"></script>
+        <script type="text/javascript" src="{% static "js/jquery.formset.min.js" %}"></script>
```
felixrindt, line 27: As far as I understand, we don't need to use the min versions here, as they get compressed anyway...

.env.example
jeriox: i understand the point for the data dir.
But I'd still prefer having the STATIC_URL and STATIC_ROOT in the .env because this is really a thing that every installation should set. you can still default this to the data dir there?

jeriox (approved):

PR #184, opened 2020-10-18
Add tests for userprofile views

gritfessel:
TODO:
- [x] test email when userprofile is created
- [ ] test qualification grants (with & without expiration)


gritfessel:
Testing qualification grants is not easily possible because initially, there are no form fields that could be filled in because of the dynamic formset. @felixrindt

felixrindt:
> Testing qualification grants is not easily possible because initially, there are no form fields that could be filled in because of the dynamic formset. @felixrindt 

You can still test them by explicitly setting post data manually, which includes values for the formset's management form. A little more work, but possible.

PR #210, opened 2020-12-21
consequences

### LINKED ISSUE #152: Consequences
I expect the system to have a model for consequences.
A consequence has a slug for the consequence type, a JSON field for additional data, a reference to the particpant it should be applied to and boolean fields for is_confirmed (default false), needs_confirmation (default false) and is_executed (default false).

I expect the system to have a model for "Helferstunde" (hour of work) with subject, start_time, end_time and an optional reference to the consequence it came from.

I expect the system to have a BaseConsequenceHandler class and a consequence handler for the `hour_of_work` consequence slug. It should offer a method to execute the consequence that creates the WorkHour object and marks the consequence as executed.

### LINKED ISSUE #155: Approve consequence
depends on #152 

As a user with the `add_consequence` permission, I expect the main menu to have a link to a "Open Consequences" page. The corresponding view should display a list of consequences with `needs_confirmation=True` and `is_confirmed=False`. I expect each consequence to offer buttons to approve or deny the consequence. The corresponding functions on the ConsequenceHandler should be called.

felixrindt:
fixes #152 fixes #155


ephios/user_management/consequences.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class WorkingHoursConsequenceHandler(BaseConsequenceHandler):
+    slug = "ephios.grant_working_hours"
+
+    @classmethod
+    def create(
+        cls,
+        user: UserProfile,
+        qualification: Qualification,
```
jeriox, line 93: why do we need the qualification here?

ephios/user_management/models.py
```diff
@@ -189,3 +189,68 @@ class QualificationGrant(Model):
 
     def __str__(self):
         return f"{self.qualification!s}, {self.user!s}"
+
+
+class Consequence(Model):
```
jeriox, line 194: this lacks the needs_confirmation field

ephios/user_management/models.py
```diff
@@ -189,3 +189,68 @@ class QualificationGrant(Model):
...
+    slug = models.CharField(max_length=255)
+    data = models.JSONField(default=dict)
+
+    user = models.ForeignKey(
+        get_user_model(), on_delete=models.CASCADE, verbose_name=_("affected user"), null=True
+    )
+    shift = models.ForeignKey(
```
jeriox, line 201: why would we want the shift? the qualification-consequence doesn't need it and the workhour-consequences saves the participation to the data dict

ephios/user_management/models.py
```diff
@@ -189,3 +189,68 @@ class QualificationGrant(Model):
...
+        "event_management.Shift",
+        on_delete=models.SET_NULL,
+        verbose_name=_("causing shift"),
+        null=True,
+    )
+
+    class States(models.IntegerChoices):
+        NEEDS_CONFIRMATION = 1, _("needs_confirmation")
+        EXECUTED = 2, _("executed")
+        FAILED = 3, _("failed")
```
jeriox, line 211: what is failed representing?

ephios/user_management/consequences.py
felixrindt: we don't, thanks

ephios/user_management/models.py
jeriox: still not sure about this

jeriox: yeah so after checking this makes no sense for working hours. so do we need it here or can we move it to the JSON? if it needs to stay here we need a blank=True for django admin

jeriox (approved):

ephios/user_management/models.py
felixrindt: It's currently used to determine who can confirm consequences...

ephios/user_management/models.py
```diff
@@ -189,3 +189,98 @@ class QualificationGrant(Model):
 
     def __str__(self):
         return f"{self.qualification!s}, {self.user!s}"
+
+    class Meta:
+        unique_together = [
+            ["qualification", "user"]
+        ]  # TODO FIX validation edge cases in UserEditForm/UserCreateForm
```
jeriox, line 196: fixme moved to #218

ephios/user_management/models.py
felixrindt: So after talking about this, there are two things to consider when deciding between the json field and a foreign key:

* prefetching objects: When rendering a list of consequences, needed objects should be prefetchable using the IDs stored in json. From what I read and tried: `prefetch_related` and `select_related` only work with actual foreign keys and don't support a pk stored in json. You could probably build a query using `extra`, but that's another topic.

* filtering consequences: you should be able to filter consequences based on data in json. That's possible with modern django.

felixrindt: I went for annotations for now. Bulks up the code quite a bit :/

PR #274, opened 2021-01-18
Create Participations in Disposition

### LINKED ISSUE #228: Add participations on disposition view
As a planner, I expect the disposition page to have controls to add participations (for participants aka users) that have not signed up for themselves. This could either be solved by a seperate area with all unrelated users that can be dragged in the same way as the signed up users or with a seperate field where I can search for the user I want to add.

### LINKED ISSUE #251: Disposition view for InstantConfirmationSignupMethod
As a planner I want to remove people that have previously been confirmed from a shift using InstantConfirmationSignupMethod. Therefor I expect to have a disposition view similar to the other signup methods

felixrindt:
Fixes #228. Also fixes #251.


felixrindt:
Also fixes #251

ephios/plugins/basesignup/signup/disposition.py
```diff
@@ -0,0 +1,150 @@
...
+        widgets = dict(state=forms.HiddenInput(attrs={"class": "state-input"}))
+
+
+def get_disposition_formset(form):
+    return forms.modelformset_factory(
+        model=AbstractParticipation,
+        form=form,
+        extra=0,
+        can_order=False,
+        can_delete=True,
```
felixrindt, line 38: Will man wirklich alle Participations löschen können? Gibts da nicht verwechslungspotential? Also leute löschen statt tatsächlich abzusagen. Wenn die created sind - ja, aber ich hab sorge dass das insb mit der federation ganz komisch ist dann.

ephios/event_management/signup.py
```diff
@@ -21,6 +21,7 @@
 
 from ephios.event_management.models import AbstractParticipation, LocalParticipation, Shift
 from ephios.extra.widgets import CustomSplitDateTimeWidget
+from ephios.plugins.basesignup.signup.disposition import BaseDispositionParticipationForm
```
felixrindt, line 24: Core code shouldn't import from plugins.

ephios/static/plugins/basesignup/js/disposition.js
```diff
@@ -0,0 +1,99 @@
...
+            participation.addClass("list-group-item-info");
+            setTimeout(() => {
+                participation.removeClass("list-group-item-info")
+            }, 2000);
+        } else {
+            // get the new form from the server
+            const addUserForm = $("#" + $(this)[0].form.id);
+            const newIndex = formset.totalFormCount();
+            addUserForm.find('[name=new_index]').val(newIndex);
+            $.ajax({
```
felixrindt, line 76: Do we need a loading indicator/disable the input while the request is running?

jeriox:
image
Is the weird vertical alignment intended?
I think I'd prefer to have the select field below the heading on desktop

ephios/plugins/basesignup/signup/disposition.py
jeriox: I'd prefer to only be able to delete participations I just created to reduce confusion. The "add others" box should also not accept drops (so that I can't move participations there) as this would also result in a deletion

ephios/static/plugins/basesignup/js/disposition.js
jeriox: i think just disabling wouldn't be a good UX. some sort of spinner would be cool, but I can't think of a nice way to place it. If you also can't come up with one now I'd also be fine with moving that to another issue.

jeriox: for me it looks like the added participation is displayed and hidden afterwards, that looks kinda strange.

ephios/event_management/signup/disposition.py
```diff
@@ -0,0 +1,167 @@
...
+                self.object.signup_method.disposition_participation_form_class
+            )
+            formset = DispositionParticipationFormset(
+                queryset=AbstractParticipation.objects.filter(pk=instance.pk),
+                prefix="participations",
+                start_index=form.cleaned_data["new_index"],
+            )
+            form = next(filter(lambda form: form.instance.id == instance.id, formset))
+            return self.render_to_response({"form": form})
+        raise Http404("User does not exist")
```
jeriox, line 129: translation?

ephios/static/ephios/js/formset/formset.js
```diff
@@ -0,0 +1,379 @@
+/**
```
jeriox, line 1: as far as I see you didn't remove the packaged version from the dependencies

jeriox (approved):

ephios/event_management/signup/disposition.py
felixrindt: this is not displayed, could remove the string entirely

ephios/static/ephios/js/formset/formset.js
felixrindt: correct, and we actually still depend on it for the template tags. Trust me I tried to remove it ;)

jeriox: sad story. did you commit your changes upstream?

felixrindt: not yet :sweat_smile:

ephios/plugins/basesignup/signup/disposition.py
felixrindt: Ok I did that. Can you try? :)

jeriox: "Deletion a participation is only allowed if it was just added through disposition."
rest is great

ephios/static/ephios/js/formset/formset.js
jeriox: okay but wth is this under ephios and not seperate like every other dependency?

PR #278, opened 2021-01-20
add django-dynamic-preferences

### LINKED ISSUE #240: Settings for event and shift defaults
As a manager I want to be able to edit the settings of the instance. This includes default values for shifts and events (e.g. who should be responsible for new events). I expect the corresponding UI elements to be hidden in the create process but accessible in an "advanced settings" section

jeriox:
closes #240 

Todo:

- [x] user notification settings for responsible things
- [x] settings on EventType (defaults for visible_for, responsible_users and responsible_groups)
- [x] hide fields on event creation
- [x] views for global settings
- [x] views for event types
- [x] fix design of "advanced settings" section for event creation/update
- [ ] write some tests (yay)


ephios/event_management/mail.py
```diff
@@ -37,19 +37,35 @@ def new_event(event):
...
 
 
 def participation_state_changed(participation: AbstractParticipation):
     messages = []
 
     # send mail to the participant whose participation has been changed
-    if participation.participant.email is not None and participation.state in (
+    mail_requested = participation.participant.email is not None
+    if participation.get_real_instance_class() == LocalParticipation:
+        local_participation = participation.get_real_instance()
```
felixrindt, line 58: my understanding was that participation already is of the correct type when using the polymorphic managers. ([Doc](https://django-polymorphic.readthedocs.io/en/stable/api/polymorphic.models.html?highlight=get_real_instance#polymorphic.models.PolymorphicModel.get_real_instance))

Is that not given here?

felixrindt (approved):
Gonna look into the css stuff...

ephios/event_management/mail.py
jeriox: just checked, this is given here

PR #310, opened 2021-02-08
CI-test on mysql and postgres

felixrindt:
Interestingly, some tests fail on those other platforms.


jeriox:
@felixrindt everything works now, but I have no earthly idea why we cannot annotate the event stuff as well. As this breaks the queryset completely instead of failing silently, I removed it. If you want to you can have another look after merging this. I approved it because it is your PR, but please review my stuff!

jeriox (approved):

PR #327, opened 2021-02-23
Better Event List

### LINKED ISSUE #235: Display participation state on event list
As a user, I expect the event list to represent my participation state for each event. This could be achieved with some color coding, badges or similar

should wait until #323

### LINKED ISSUE #277: Improve event list
The event list has some room for improvement. Ideas:

- we could use the grid system instead of a table for the entries, this would allow for a more flexible layout
- it would be nice to display the additional information (end date, type etc.) on small screens as well
- it would be nice to display (some) shifts of each event and/or the times

### LINKED ISSUE #323: Add "minimum amount of users" to InstantConfirmation and RequestConfirm
As a planner
When creating a shift with InstantConfirmation or RequestConfirm signup method
I expect the form to have a field where i can enter a number for a minimum amount of participants
So that the system knows if there are volunteers missing for this shift

felixrindt:
fixes #235 fixes #277 fixes #323

* [x] Center action button
* [ ] event type label colors -> new issue
* [x] the whole line as touch target

### DIFF
--- a/ephios/core/templatetags/event_extras.py
+++ b/ephios/core/templatetags/event_extras.py
@@ -67,7 +67,7 @@ def event_signup_state_counts(event, user):
     counter = collections.Counter()
     for shift in event.shifts.all():
         for participation in shift.participations.all():
-            if type(participation) is LocalParticipation and participation.user_id == user.id:
+            if isinstance(participation, LocalParticipation) and participation.user_id == user.id:
                 counter[participation.state] += 1
                 break
     return counter
--- a/ephios/core/templatetags/event_extras.py
+++ b/ephios/core/templatetags/event_extras.py
@@ -3,7 +3,7 @@ from datetime import datetime
 from django import template
 from django.utils.safestring import mark_safe
 
-from ephios.core.models import AbstractParticipation
+from ephios.core.models import AbstractParticipation, LocalParticipation
 
 register = template.Library()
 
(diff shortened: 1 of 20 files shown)


ephios/core/templatetags/event_extras.py
```diff
@@ -55,3 +56,18 @@ def confirmed_shifts(user):
...
+
+
+@register.filter(name="event_signup_state_counts")
+def event_signup_state_counts(event, user):
+    """
+    Return a counter counting states for a users participations for shifts of an event.
+    Uses None as key for no participation info.
+    """
+    counter = collections.Counter()
+    for shift in event.shifts.all():
```
jeriox, line 68: is there any chance of caching smth here? this feels like a very big loop over everything

ephios/extra/templatetags/utils.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@register.filter(name="get")
+def dict_get(dictionary, key):
+    return dictionary.get(key)
+
+
+@register.filter(name="sum")
+def _sum(elements):
+    print(elements)
```
jeriox, line 13: print statement

ephios/static/ephios/js/main.js
```diff
@@ -31,6 +31,13 @@ $(document).ready(function () {
             $('#checkall').prop('checked', false);
         }
     });
+
+    $("[data-jslink]").on("click", function (ev) {
```
jeriox, line 35: i'm not sure if I like this solution, mainly because context actions don't work on mobile (so I cannot hold the event and select "open in a new tab" and I hate it when web apps do that)

jeriox (changes requested):

jeriox:
oh and maybe a short test on the SignupStats functionality

felixrindt:
Linter fails because of https://github.com/PyCQA/pylint/issues/3882

jeriox (changes requested):
- in Chrome i can't select text. If I double-click an event in the list, the text of the footer is selected.
- on mobile the checkbox isn't aligned with the user icon anymore

felixrindt:
couldn't do something about text selection on chrome

jeriox (approved):

PR #335, opened 2021-02-28
Password reset button for managers

### LINKED ISSUE #287: Password Reset Button for Managers
As a manager,  
I want to be able to click a button on user list or in the user form to quickly send a password reset link to the user  
so that I don't need to go the long route and copy-paste the email into the password reset form.

gritfessel:
closes #287


ephios/core/templates/core/userprofile_confirm_password_reset.html
```diff
@@ -0,0 +1,22 @@
...
+{% block title %}
+    {% translate "Password reset" %}
+{% endblock %}
+
+{% block content %}
+    <div class="page-header">
+        <h1>{% translate "Password reset" %}</h1>
+    </div>
+    <p>{% blocktranslate with email=userprofile.email %}
+        Are you sure that you want to reset the password of the user "{{ email }}"?
```
jeriox, line 13: i think I'd prefer having the full name instead of the email here

ephios/core/templates/core/userprofile_confirm_password_reset.html
```diff
@@ -0,0 +1,22 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block title %}
+    {% translate "Password reset" %}
+{% endblock %}
+
+{% block content %}
+    <div class="page-header">
+        <h1>{% translate "Password reset" %}</h1>
```
jeriox, line 10: maybe indicate that this is meant for another user to distinguish from the own password reset?

ephios/core/views/accounts.py
```diff
@@ -136,6 +137,49 @@ def get_success_url(self):
...
+    model = UserProfile
+    permission_required = "core.change_userprofile"
+    template_name = "core/userprofile_confirm_password_reset.html"
+
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        self.object = self.get_object()
+
+    def post(self, request, *args, **kwargs):
+        userprofile = self.object
```
jeriox, line 150: this looks duplicated

ephios/core/views/accounts.py
```diff
@@ -136,6 +137,49 @@ def get_success_url(self):
...
+        if not userprofile:
```
jeriox, line 151: SingleObjectMixin already raises Http404 if the object is not found

ephios/core/templates/core/userprofile_form.html
```diff
@@ -19,6 +19,13 @@ <h1>{% trans "Edit user" %}</h1>
             <h1>{% trans "Add new user" %}</h1>
         {% endif %}
     </div>
+    {% if userprofile %}
+        <div class="float-right d-flex">
```
jeriox, line 23: I don't like that the button is so close to the field below and it appears kinda lost up there. Maybe put it next to delete button which triggers a "similar" action? Also I'd prefer to have another color for the button as this now indicates the same destructivity as deleting the user.

jeriox (changes requested):

ephios/core/views/accounts.py
gritfessel: what exactly?

ephios/core/templates/core/userprofile_form.html
gritfessel: grafik
I feel like this looks also weird... more ideas? I guess it would be okay though

jeriox (approved):

PR #337, opened 2021-02-28
Add display of pending consequences for the current user

### LINKED ISSUE #239: Pending consequences
As a user, I expect the homepage to list my pending consequences (state needs_confirmation). I also expect to receive an email once my consequence has been executed or declined.

### LINKED ISSUE #262: Improve decimal formatting for consequences/workhours
Currently, the consequence render() function rounds to one digit and point separator and the profile rounds to one digit with comma separator.

We want the consequence render to use _up to_ 2 digits with the appropriate local seperator (use django) and the profile view to use exactly two digits with the appropriate local separator.

strawberriesandcheese:
Closes #239 closes #262


ephios/core/consequences.py
```diff
@@ -43,6 +43,11 @@ def editable_consequences(user):
     return qs.filter(slug__in=map(lambda hl: hl.slug, handlers)).distinct()
 
 
+def my_pending_consequences(user):
```
jeriox, line 46: it is called "my" consequences but can be given any user. maybe just pending consequences?

strawberriesandcheese: The problem is that "pending consequences" are all consequences that are pending. Here we are just talking about the pending consequences of a sepcific user tho. I think it would make the code less readable when using this method if it wss only called pending consequences. @felixrindt wanna also defend this? :D

felixrindt:
```suggestion
def pending_consequences_acting_on(user):
```

felixrindt: Well I guess that doesn't work for a filter but well

ephios/core/templatetags/user_extras.py
```diff
@@ -18,6 +18,12 @@ def editable_consequences_tag(user, states=None):
     return qs
 
 
+@register.filter(name="pending_consequences")
+def pending_consequences_filter(user):
```
felixrindt, line 22:
```suggestion
register.filter(name="pending_consequences")(pending_consequences)
```

felixrindt: should suffice?! (no need to build an id function)
Maybe there also is nicer syntax to register existing syntax as filters.

felixrindt (commented):
Maybe we can increase coverage?!

felixrindt (approved):

PR #338, opened 2021-02-28
use rich_text for event description

### LINKED ISSUE #313: Format event description
As a planner, I want to be able to format the event description. This should at least mean that links will be formatted so that they are clickable. Maybe this should mean that the description will accept Markdown.

jeriox:
close #313

### DIFF
--- a/ephios/extra/templatetags/rich_text.py
+++ b/ephios/extra/templatetags/rich_text.py
@@ -15,6 +15,8 @@ ALLOWED_TAGS = [
     "code",
     "div",
     "em",
+    "h1",
+    "h2",
     "h3",
     "h4",
     "h5",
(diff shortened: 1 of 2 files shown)


ephios/extra/templatetags/rich_text.py
```diff
@@ -15,8 +15,6 @@
     "code",
     "div",
     "em",
-    "h1",
-    "h2",
```
felixrindt, line 19: Well actually we want these headings for pages.

felixrindt: Can we make not having them optional?

felixrindt (changes requested):

ephios/extra/templatetags/rich_text.py
jeriox: do we? as the title is already rendered as h1 i'd prefer to not have h1 in the content there.

felixrindt: I'd opt for having them. Not having them would make `rich_text` really unintuitive.

felixrindt (approved):

PR #343, opened 2021-03-02
display qualifications in userlist

### LINKED ISSUE #252: Display qualifications on DispositionView
As a planner I expect to be informed about the participants qualification while dispatching. This could be implemented by some labels on the participant fragment. 
I expect the system to have a setting where I can decide which qualification categories should be displayed as such labels.

### LINKED ISSUE #273: Additional columns in user list
As a manager, I want to have a setting where I can select a set of QualificationCategories. I expect the user list to display an additional column for each of those categories with the highest qualification each user has in that category.
This shares functionality with #252, maybe it could even use the same setting.

jeriox:
closes #273 closes #252

### DIFF
--- a/ephios/core/views/accounts.py
+++ b/ephios/core/views/accounts.py
@@ -2,6 +2,7 @@ from django.contrib import messages
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.auth.models import Group
 from django.contrib.messages.views import SuccessMessageMixin
+from django.db.models import Prefetch
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.utils.translation import gettext as _
@@ -19,7 +20,7 @@ from dynamic_preferences.users.views import UserPreferenceFormView
 
 from ephios.core import mail
 from ephios.core.forms.users import GroupForm, QualificationGrantFormset, UserProfileForm
-from ephios.core.models import UserProfile
+from ephios.core.models import QualificationGrant, UserProfile
 from ephios.extra.permissions import CustomPermissionRequiredMixin
 
 
@@ -37,10 +38,15 @@ class UserProfileListView(CustomPermissionRequiredMixin, ListView):
         global_preferences = global_preferences_registry.manager()
         categories = global_preferences["general__relevant_qualification_categories"]
         qs = UserProfile.objects.all()
-        qs2 = qs.prefetch_related("qualification_grants").prefetch_related(
-            "qualification_grants__qualification"
-        )
-        return qs2
+        for category in categories:
+            qs = qs.prefetch_related(
+                Prefetch(
+                    "qualification_grants",
+                    queryset=QualificationGrant.objects.filter(qualification__category=category),
+                    to_attr=str(category.pk),
+                )
+            ).prefetch_related(f"{category.pk}__qualification")
+        return qs
 
 
 class UserProfileCreateView(CustomPermissionRequiredMixin, TemplateView):
--- a/ephios/core/views/accounts.py
+++ b/ephios/core/views/accounts.py
@@ -14,6 +14,7 @@ from django.views.generic import (
     UpdateView,
 )
 from django.views.generic.detail import SingleObjectMixin
+from dynamic_preferences.registries import global_preferences_registry
 from dynamic_preferences.users.views import UserPreferenceFormView
 
 from ephios.core import mail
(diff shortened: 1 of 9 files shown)


ephios/core/views/accounts.py
```diff
@@ -32,6 +34,20 @@ class UserProfileListView(CustomPermissionRequiredMixin, ListView):
...
+        categories = global_preferences["general__relevant_qualification_categories"]
+        qs = UserProfile.objects.all()
+        for category in categories:
+            qs = qs.prefetch_related(
+                Prefetch(
+                    "qualification_grants",
+                    queryset=QualificationGrant.objects.filter(qualification__category=category),
+                    to_attr=str(category.pk),
+                )
+            ).prefetch_related(f"{category.pk}__qualification")
```
felixrindt, line 48: That last one should be a select related going from n:1 (grant:qualification)

ephios/core/views/accounts.py
```diff
@@ -32,6 +34,20 @@ class UserProfileListView(CustomPermissionRequiredMixin, ListView):
...
+    def get_queryset(self):
+        global_preferences = global_preferences_registry.manager()
+        categories = global_preferences["general__relevant_qualification_categories"]
+        qs = UserProfile.objects.all()
+        for category in categories:
+            qs = qs.prefetch_related(
+                Prefetch(
+                    "qualification_grants",
+                    queryset=QualificationGrant.objects.filter(qualification__category=category),
+                    to_attr=str(category.pk),
```
felixrindt, line 46: Can we do something more verbose than integer ids? Like `qualification_category_{id}`?

ephios/plugins/basesignup/templates/basesignup/section_based/fragment_participant.html
```diff
@@ -1,20 +1,28 @@
...
+                    {% for qualification in form.instance.participant.qualifications|get_relevant_qualifications %}
+                        <span class="badge badge-secondary">{{ qualification }}</span>
+                    {% endfor %}
+                </div>
             </div>
-            <div class="col-6 d-flex justify-content-between align-items-center">
+            <div class="col-4 d-flex justify-content-between align-items-center">
                 {% bootstrap_field form.section show_label='sr-only' form_group_class="form-group d-inline-block mb-0" %}
 
                 {% if form.can_delete %}
```
felixrindt, line 28: damn github I can't mark line 29 (new). Can we do `ml-1` instead of `ml-3` here? Space is a little wasted otherwise...

image

felixrindt (changes requested):

ephios/plugins/basesignup/templates/basesignup/section_based/fragment_participant.html
jeriox: this is not my code but yeah sure ^^

ephios/core/views/accounts.py
jeriox: it's just a bit easier to access in the template because stuffing strings together there is always such a pain

ephios/core/views/accounts.py
jeriox: i tried that, select_related seems not be able to use prefetched stuff, it throws a field error with `Invalid field name(s) given in select_related: '2', '1'. Choices are: (none)`

felixrindt: Did you try putting it on the queryset given to the Prefetch-object? I think that's what the documentation shows.

ephios/plugins/basesignup/templates/basesignup/section_based/fragment_participant.html
felixrindt: Well you did make the columns smaller, didn't ya? :p

ephios/core/views/accounts.py
jeriox: i used 
```
qs = qs.prefetch_related(
                Prefetch(
                    "qualification_grants",
                    queryset=QualificationGrant.objects.filter(qualification__category=category),
                    to_attr=f"qualifications_for_category_{category.pk}",
                )
            ).select_related(f"qualifications_for_category_{category.pk}__qualification")
```
I dont't think that there is another possible application order as the field is "created" by the first prefetch

felixrindt: Here's what I meant:

```python
qs = qs.prefetch_related(
                Prefetch(
                    "qualification_grants",
                    queryset=QualificationGrant.objects.filter(
                        qualification__category=category,
                    ).select_related(f"qualifications_for_category_{category.pk}__qualification"),
                    to_attr=f"qualifications_for_category_{category.pk}",
                )
            )
```

jeriox: crazy shit :D

felixrindt: :D

felixrindt (approved):
I still don't know how to feel about the global config of "important" qualifications but it's probably good for now.

jeriox:
> I still don't know how to feel about the global config of "important" qualifications but it's probably good for now.

yeah i also came across that thought. I think we'll stick to it for now until somebody wants to differentiate between the places where we use it...

PR #345, opened 2021-03-02
display shift status in pdf export

### LINKED ISSUE #88: improve PDF export for SectionBasedSignupMethod
As a planner, I'd like to see the positions and their required qualifications in the PDF export. If the position is already taken, there should be the name of the user next to it. See the PDF linked in #10 for reference. Probably we want the signup methods to deliver something like a dict that is rendered to the table in the PDF file.

jeriox:
closes #88 

TODO
- [x] remove helper code once #343 is merged
- [x] figure out a way to use `get_participation_display()` of InstantConfirmationSignupMethod on RequestConfirmSignupMethod
- [x] improve layout (padding of table, borders)
- [x] render sections with no participants


ephios/plugins/basesignup/signup/common.py
```diff
@@ -102,3 +103,36 @@ def get_configuration_fields(self):
...
+                ", ".join(
+                    participant.qualifications.filter(
+                        category__in=relevant_qualification_categories
+                    ).values_list("title", flat=True)
+                ),
+            ]
+            for participant in self.shift.get_participants()
+        ]
+        try:
+            # we may not be using MinMaxParticipantsMixin so there may be no bounds
```
felixrindt, line 123: Well that comment is not completely right. The Mixin only provides one way to implement `get_participant_count_bounds`. SectionBasedSignup doesn't use it and still implements the method.

ephios/core/signup/methods.py
```diff
@@ -378,6 +378,15 @@ def render_shift_state(self, request):
         """
         return ""
 
+    def get_participation_display(self):
+        """
+        Returns a displayable representation of participation (list of lists) that can be rendered into a table.
+        """
```
felixrindt, line 384: As I see it, this data will mainly be used for pdfs, what else? Could better reflect that in the method name.
Also, I'd like to see the docstring go into more detail about the format and data to include.

ephios/plugins/basesignup/signup/common.py
felixrindt: And plugins might use the QualificationMixin but not the MinMaxMixin but their own way...

ephios/core/signup/methods.py
jeriox: yeah i wasn't really sure about how the separation should be done here. I started by building the section of the PDF completely in the signup method, but this isn't perfect as we don't know to which pagesize the data will be rendered. so I switched over to this approach where the method only delivers a list that *could* be used elsewhere in the future (say some kind of dashboard or whatever). this is why I didn't name the method specifically for pdf usage

felixrindt: If you could elaborate on the "lists of lists" a little in the docstring, that's still be great.

felixrindt (approved):

ephios/core/signup/methods.py
felixrindt: <3

PR #346, opened 2021-03-04
add event notification view

### LINKED ISSUE #229: Send reminder via mail
As a planner, I expect the event detail page to have a button to resend a notification to users. I only expect users that have no participations for that event to receive the mail.

jeriox:
closes #229


ephios/core/forms/events.py
```diff
@@ -207,3 +209,36 @@ def render(self):
 
     def save(self):
         pass
+
+
+class EventNotificationForm(forms.Form):
+    NEW_EVENT = "new"
+    REMINDER = "remind"
+    PARTICIPANTS = "participants"
```
felixrindt, line 217: Elsewhere we are using `models.TextChoices`. Feel free to do that here as well.

ephios/static/ephios/css/main.css
```diff
@@ -83,3 +83,7 @@ This hack fixes that. Inspired by https://github.com/select2/select2/issues/4220
 .select2 {
     width: 100% !important;
 }
+
+.no-display {
+    display: none;
+}
```
felixrindt, line 89: Why not use `d-none` from bootstrap?

ephios/static/ephios/js/event_notifications.js
```diff
@@ -0,0 +1,9 @@
+$(document).ready(function () {
+    $("input:radio[name ='action']").change(function (event) {
+        if ($("input:radio[name ='action']:checked").val() === "participants")
+            $("#div_id_mail_content").slideDown();
+        else
+            $("#div_id_mail_content").slideUp();
+    });
+
```
felixrindt, line 8: I'm not a big fan of adding this `show-if-this-is-checked` logic in multiple views, we could add something like this as a general purpose utility: 
https://github.com/pretix/pretix/blob/a897f60fc5a22d9eb9bf4107111eb8ea05d09b3b/src/pretix/static/pretixcontrol/js/ui/main.js#L287

felixrindt (approved):
I'm approving this so we can move on, but please have a look at the comments. Esp. with the JS-thing I don't want to have dozens of page-specific js files flying around.

ephios/core/forms/events.py
felixrindt: or IntegerChoice for that matter, just FYI

ephios/static/ephios/css/main.css
jeriox: because the boostrap one uses !important and the jquery slideDown does not override the !important flag, so the content is not shown

ephios/core/forms/events.py
jeriox: well but we are are form and we don't have a model

PR #348, opened 2021-03-04
Fix button tooltips

felixrindt:
I guess it wasn't broken on small screens :sweat_smile:


jeriox:
image
sometimes the tooltip contains divs and sometimes ps which leads to a bit different styling

jeriox (approved):

PR #352, opened 2021-03-05
Guest Participations

felixrindt:
This PR does some groundwork for non-local participants and adds a guest participation plugin.

Try it by going to: http://localhost:8000/guests/register/1/secret/

I do not plan on making this feature complete until we have users that actually need it.

Possible features:
* better privacy: hide participant names for unauthenticated users
* plugin hook for event copy
* export of emails / a way to communicate information to guests
* document upload (e.g. proof for qualifications)


ephios/core/signup/methods.py
```diff
@@ -78,6 +83,16 @@ def collect_all_qualifications(self) -> set:
...
 
+    def reverse_signup_action(self, shift):
+        raise NotImplementedError
+
+    def reverse_event_detail(self, event):
+        raise NotImplementedError
+
+    @property
+    def icon(self):
+        return mark_safe('<span class="fa fa-user"></span>')
```
felixrindt, line 94: I'd love to avoid adding more obligatory methods to the AbstractParticipant interface.
But we need some way for Signupviews to know where to redirect users, because I don't think we can easily have the same view, e.g. the EventDetailView implement multiple ways of authenticating (and also there might be different url kwargs, etc...), hence I know doubled those views.

ephios/core/signup/methods.py
```diff
@@ -92,15 +107,51 @@ def participation_for(self, shift):
...
 class ParticipationError(ValidationError):
     pass
 
 
+def prevent_getting_participant_from_request_user(request):
+    original_user = request.user
+
+    class ProtectedUser(SimpleLazyObject):
+        def as_participant(self):
+            raise Exception("Access of request.user.as_participant in SignupViews is not allowed.")
```
felixrindt, line 126: I tried all kinds of mechanics to enforce this in some kind of way, trying context managers, overwriting instance properties. This is what worked best. 
Of course we can replace the exception with a warning, but it works so far.

ephios/core/signup/methods.py
felixrindt: What about adding another signal that collects a set of participationtypes including versions of the view? I can't think of a nice architecture right now.

felixrindt:
I guess it's now mostly functional.

ephios/core/signals.py
```diff
@@ -33,6 +33,19 @@
...
+"""
+This signal is sent out to get a participant from a request with an unauthenticated user.
+Return a subclass of AbstractParticipant or None if you cannot provide a participant.
+The first non-None return-value will be used.
+Receivers will receive a ``request`` keyword argument.
+"""
+
+event_forms = PluginSignal()
+"""
+This signal...
```
jeriox, line 46: ...?

ephios/core/views/event.py
```diff
@@ -74,13 +98,27 @@ def get_form(self, form_class=None):
             data=self.request.POST or None, user=self.request.user, instance=self.object
         )
 
+    def post(self, request, *args, **kwargs):
```
jeriox, line 101: why is this necessary?

ephios/plugins/guests/templates/guests/event_detail.html
```diff
@@ -0,0 +1,10 @@
+{% extends "core/event_detail.html" %}
+{% load bootstrap %}
+{% load rich_text %}
+{% load guardian_tags %}
+{% load i18n %}
+
+{% block content %}
+    <h3 class="text-info">{{ object.as_participant }}</h3>
```
jeriox, line 8: this seems a bit weird with just the name as such a big heading

ephios/plugins/guests/views.py
```diff
@@ -0,0 +1,104 @@
...
+    def get_authenticated_url(self):
+        return reverse("core:index")
+
+
+class GuestRegistrationView(RedirectAuthenticatedUserMixin, CreateView):
+    model = GuestUser
+    form_class = forms.modelform_factory(
+        GuestUser,
+        fields=["email", "first_name", "last_name", "date_of_birth", "phone", "qualifications"],
+        widgets={"qualifications": Select2MultipleWidget},
```
jeriox, line 36: DOB please with our custom date widget

ephios/plugins/guests/templates/guests/guestuser_detail.html
```diff
@@ -0,0 +1,28 @@
+{% extends "base.html" %}
```
jeriox, line 1: I can't see where this template is used?

jeriox:
image
when checking the "allow guests" checkbox on an existing event and saving, this happens for me

jeriox:
> image
> when checking the "allow guests" checkbox on an existing event and saving, this happens for me

works flawlessly for our debug event but not for a custom event with one shift with requestconfirm

ephios/plugins/guests/views.py
```diff
@@ -0,0 +1,104 @@
...
+    def form_valid(self, form):
+        form.instance.event = self.event
+        try:
+            guest = form.save()
+        except IntegrityError:
+            # unique_together constraint not passed
+            form.add_error(
+                None,
+                _(
+                    "You already registered as a guest for this event. Check your email to access the event."
```
jeriox, line 71: you wanna actually send the mail if the message says that it is sent? 🙈

ephios/core/signals.py
felixrindt: woops

ephios/plugins/guests/templates/guests/event_detail.html
felixrindt: I can make it smaller :D
Was just simple this way...

.github/workflows/staging.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         key: ${{secrets.STAGING_SSH_PRIVATE_KEY}}
         inventory: |
           [servers]
-          despina.uberspace.de ansible_user=staghios domain=staging.ephios.de smtp_mailbox=noreply admin_email=admin@ephios.de
+          despina.uberspace.de ansible_user=staghios domain=staging.ephios.de smtp_mailbox=noreply admin_email="Ephios Admins <admin@ephios.de>"
```
felixrindt, line 63: I hope this works? Didn't get emails from staging. Although I still find this format unintuitive.

jeriox: yes this should be correct.
well this seems to be some kind of standard format, at least `from email.utils import getaddresses` is used, so this comes from python directly

jeriox (approved):

PR #354, opened 2021-03-07
refactor permission management in forms

### LINKED ISSUE #299: Refactor role/permission system
* Staff user
    * access to Django-Admin, Instanz-Einstellungen (API-Keys, Name, Logo, ...)
    * using plain Django-Staff 
* Gruppen-Rollen
    * Management-Rolle: 
        * CRUD Gruppen, Eventtypen, Qualifaktionen, Qualifikationskategorien
        * CRUD alle Events
        * CRUD alle Nutzerprofile
        * alle Gruppenmitgliedschaften bearbeiten
    * Benutzerverwaltungs-Rolle: 
        * CRUD alle Nutzerprofile
        * Gruppenmitgliedschaften für Gruppen, denen der aktuelle User angehört
    * Planer-Rolle: 
        * Kann Events hinzufügen (für bestimmte Gruppen)

jeriox:
close #299


ephios/static/ephios/js/group_form.js
```diff
@@ -1,11 +1,31 @@
...
+            $("#id_is_hr_group").prop("checked", true);
+            $("#id_is_planning_group").prop("disabled", true);
+            $("#id_is_planning_group").prop("checked", true);
+            $(".publish-select").slideDown();
+        }
+        else {
+            $("#id_is_hr_group").prop("disabled", false);
+            $("#id_is_hr_group").prop("checked", false);
+            $("#id_is_planning_group").prop("disabled", false);
+            $("#id_is_planning_group").prop("checked", false);
```
felixrindt, line 23: I think we should leave the two checkboxes checked.

felixrindt: or set them to their initial value when available

ephios/static/ephios/js/group_form.js
```diff
@@ -1,11 +1,31 @@
...
+            $(".publish-select").slideUp();
```
felixrindt, line 24: can we have one call for the slideUp/Down at the end of the function based on the state of `is_planning_group`?

felixrindt: and one single handler for all checkboxes? I think it would be nicer:

```javascript
function handleGroupForm(event) {
    // do the sliding and checking based on :checked
    // do the sliding instantly if event is null
}

$(() => {
    handleGroupForm();
    $(".groupformcheckbox").click(handleGroupForm);
});
```

ephios/core/forms/users.py
```diff
@@ -20,43 +20,82 @@
...
+        ],
         required=False,
     )
-    can_manage_group = BooleanField(
-        label=_("Can manage groups"),
+    is_management_group = PermissionField(
+        label=_("Can manage ephios"),
         help_text=_(
-            "If checked, users in this group can add and edit all groups, their permissions as well as group memberships."
+            "If checked, users in this group can manage users, groups, eventtypes and qualifications"
```
felixrindt, line 69: include "can change all group memberships"

ephios/core/forms/users.py
```diff
@@ -20,43 +20,82 @@
...
+        required=False,
+        help_text=_(
+            "Choose groups that the group you are currently editing can decide whether to grant working hours for."
+        ),
+        widget=Select2MultipleWidget,
+    )
+
+    is_hr_group = PermissionField(
+        label=_("Can edit users"),
         help_text=_("If checked, users in this group can view, add, edit and delete users."),
```
felixrindt, line 57: add "users in this group can also manage group memberships regarding any of their own groups." or smth like that.

felixrindt (changes requested):

ephios/extra/mixins.py
```diff
@@ -119,3 +121,18 @@ def get_canonical_slug(self):
         object class. In that case, this method will never be called.
         """
         return self.get_object().slug
+
+
+class PermissionFormMixin:
```
felixrindt, line 126: Can this be in the same file as the corresponding field and get a docstring that they work together please?

ephios/extra/mixins.py
```diff
@@ -119,3 +121,18 @@ def get_canonical_slug(self):
...
+        for field in self.fields.values():
+            if isinstance(field, PermissionField) and self.instance.pk is not None:
+                field.set_initial_value(self.permission_target)
+
+    def save(self, commit=True):
+        target = super().save(commit)
+        for key, field in self.fields.items():
+            if isinstance(field, PermissionField) and key in self.changed_data:
+                field.update_permissions(target, self.cleaned_data[key])
+        return target
```
felixrindt, line 138: As predicted we are getting in trouble here because the permission sets aren't mutually exclusive. Reproduce:

Add a non-staff user to a group. Make that group a management group. Save the form.

In a second form cycle, remove the management permission but leave the hr_management checkbox checked.

The user will then not be able to edit users, even though their group had the HR checkbox checked. When reloading the form, the checkbox will be unchecked.

Here's a possible fix:

1. Don't restrict to `self.changed_data`. Unchanged fields could still provide permissions important because of step 2.
2. First, update permissions for all unchecked Permission fields. Then add all permissions from checked permission fields.

ephios/extra/permissions.py
```diff
@@ -26,3 +28,23 @@ def get_groups_with_perms(obj, only_with_perms_in):
...
+            codename__in=map(lambda perm: perm.split(".")[-1], self.permission_set)
+        ).count() == len(self.permission_set)
+
+    def update_permissions(self, target, assign):
+        if assign:
+            for permission in self.permission_set:
+                assign_perm(permission, target)
+        else:
+            for permission in self.permission_set:
+                remove_perm(permission, target)
```
felixrindt, line 50: Some of the added lines are not covered by tests. Please try to cover all lines.

ephios/extra/permissions.py
```diff
@@ -26,3 +28,23 @@ def get_groups_with_perms(obj, only_with_perms_in):
...
+class PermissionField(BooleanField):
+    def __init__(self, *args, **kwargs):
+        self.permission_set = kwargs.pop("permissions")
+        super().__init__(*args, **kwargs)
+
+    def set_initial_value(self, user_or_group):
+        self.target = user_or_group
+        self.initial = self.target.permissions.filter(
+            codename__in=map(lambda perm: perm.split(".")[-1], self.permission_set)
+        ).count() == len(self.permission_set)
```
felixrindt, line 42: This feels a little hacky. Can you please try to

1. Also check for the permission `app_label` if available?
2. Actually compare the sets and not the size of it? I know they aren't of the same types atm. Maybe guardian has a function to help us here.

ephios/core/forms/users.py
```diff
@@ -66,90 +105,48 @@ def __init__(self, **kwargs):
...
+            Fieldset(
+                _("Planning"),
+                Field(
+                    "is_planning_group",
+                    title="This permission is included with the management role.",
+                ),
+                Field("publish_event_for_group", wrapper_class="publish-select"),
+                "decide_workinghours_for_group",
+            ),
+            FormActions(Submit("submit", _("Save"))),
```
felixrindt, line 137: I think this looks really clean rendered. You are really improving the permission settings IMO! :)

felixrindt (changes requested):
It's a hassle, but I find this really improves the permission situation. :rocket:

ephios/extra/mixins.py
felixrindt: If we go that route, we should also split `update_permissions` into `assign_perms` and `remove_perms` to indicate that these are two different and exclusive phases in the PermissionField Livecycle.

ephios/static/ephios/js/group_form.js
jeriox: hm i think i like it this way. when editing a group with management permissions and revoking the management permission it feels more natural to revoke the included permissions as well

ephios/extra/mixins.py
jeriox: it has been that way, i thought you'd prefer having all mixins in mixins.py ^^

ephios/static/ephios/js/group_form.js
jeriox: i don't understand what i should do here

ephios/static/ephios/js/group_form.js
felixrindt: I didn't think of the other being "included" but being obligatory when being a manager, and I still believe that's the more sensible way of thinking about it when showing them all as seperate checkboxes.

ephios/static/ephios/js/group_form.js
felixrindt: No problem, have a look at #360

ephios/static/ephios/js/group_form.js
jeriox: but I never actually checked them, so why do I need to uncheck them once I want to revoke the management permission?

felixrindt: There probably is no right answer here. Well maybe you did check them, maybe you didn't. I'd say the more common use case is to only revoke some permissions, not all of them at once.

ephios/extra/permissions.py
jeriox: Cannot resolve keyword 'app_label' into field. Choices are: codename, content_type, content_type_id, group, groupobjectpermission, id, name, user, userobjectpermission

felixrindt: ah well I wasn't sure :see_no_evil:

ephios/extra/permissions.py
```diff
@@ -43,9 +43,15 @@ def __init__(self, *args, **kwargs):
...
-        ).count() == len(self.permission_set)
+        codename_set = set(map(lambda perm: perm.split(".")[-1], self.permission_set))
+        self.initial = (
+            set(
+                self.target.permissions.filter(codename__in=codename_set).values_list(
+                    "codename", flat=True
+                )
+            )
+            == codename_set
+        )
```
felixrindt, line 54: Atlernative: `codename_set <= set(self.target.permissions).values_list("codename", flat=True))`

ephios/extra/permissions.py
```diff
@@ -75,7 +75,16 @@ def __init__(self, *args, **kwargs):
 
     def save(self, commit=True):
         target = super().save(commit)
+        to_remove = set()
+        to_assign = set()
```
felixrindt, line 79: Sets are nice, but I would have expected lists :joy:

felixrindt (approved):

PR #361, opened 2021-03-09
add conflicting_shifts signup checker

### LINKED ISSUE #233: Add "conflicting shifts" signup checker to BaseSignupMethod
The base signup method should also check whether there are any conflicting shifts for that user (overlapping times)

jeriox:
closes #233

### DIFF
--- a/ephios/core/templates/core/disposition/fragment_participation.html
+++ b/ephios/core/templates/core/disposition/fragment_participation.html
@@ -8,6 +8,13 @@
             <div class="col-10 pr-3">
                 {{ form.instance.participant.icon }}
                 {{ form.instance.participant }}
+                {% with form.instance.participant|conflicting_shifts:shift as conflicting_shifts %}
+                    {% if conflicting_shifts %}
+                        <span class="fa fa-exclamation-triangle text-danger"
+                              data-toggle="tooltip"
+                              title="{{ form.instance.participant }} already participates in {{ conflicting_shifts|join:", " }} at the same time."></span>
+                    {% endif %}
+                {% endwith %}
                 {% for qualification in form.instance.participant.qualifications|get_relevant_qualifications %}
                     <span class="badge badge-secondary">{{ qualification }}</span>
                 {% endfor %}
--- a/ephios/plugins/basesignup/templates/basesignup/section_based/fragment_participant.html
+++ b/ephios/plugins/basesignup/templates/basesignup/section_based/fragment_participant.html
@@ -17,6 +17,13 @@
                     {% endif %}
                 </div>
                 <div>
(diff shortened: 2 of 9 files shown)


ephios/plugins/basesignup/templates/basesignup/section_based/fragment_participant.html
```diff
@@ -17,6 +17,13 @@ <h6 class="mb-0">
                     {% endif %}
                 </div>
                 <div>
+                    {% with form.instance.participant|conflicting_shifts:shift as conflicting_shifts %}
+                        {% if conflicting_shifts %}
+                            <span class="fa fa-exclamation-triangle text-danger"
+                                  data-toggle="tooltip"
+                                  title="{{ form.instance.participant }} already participates in {{ conflicting_shifts|join:", " }} at the same time."></span>
+                        {% endif %}
+                    {% endwith %}
```
felixrindt, line 26: We still need to merge that with the section signup participant fragment...

felixrindt (changes requested):

ephios/plugins/basesignup/templates/basesignup/section_based/fragment_participant.html
jeriox: not sure if that is actually possible as they use a different base layout (col-8 vs col-10)

jeriox: also section_based renders the desired section below the participants name which means that just adding another column is not enough

felixrindt: ah alright.

ephios/core/templates/core/disposition/fragment_participation.html
```diff
@@ -8,6 +8,13 @@ <h6 class="mb-0">
             <div class="col-10 pr-3">
                 {{ form.instance.participant.icon }}
                 {{ form.instance.participant }}
+                {% with form.instance.participant|conflicting_shifts:shift as conflicting_shifts %}
+                    {% if conflicting_shifts %}
+                        <span class="fa fa-exclamation-triangle text-danger"
+                              data-toggle="tooltip"
+                              title="{{ form.instance.participant }} already participates in {{ conflicting_shifts|join:", " }} at the same time."></span>
+                    {% endif %}
+                {% endwith %}
```
felixrindt, line 17: You should probably call `handleForms` in the `disposition.js` when adding new forms to get the tooltip displayed correctly.

felixrindt (changes requested):
So simple but such a nice feature!

felixrindt (approved):

PR #364, opened 2021-03-10
add redis caching

### LINKED ISSUE #298: Add caching
As of #278 some settings reside in the database, so we should defintely cache database requests.

jeriox:
together with https://github.com/ephios-dev/ephios-ansible/pull/2
closes #298

### DIFF
--- a/ephios/settings.py
+++ b/ephios/settings.py
@@ -114,6 +114,9 @@ WSGI_APPLICATION = "ephios.wsgi.application"
 
 DATABASES = {"default": env.db_url()}
 
+# Caches
+CACHES = {"default": env.cache_url(default="locmemcache://")}
+
 # Password validation
 # https://docs.djangoproject.com/en/3.0/ref/settings/#auth-password-validators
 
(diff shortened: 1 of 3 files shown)


ephios/settings.py
```diff
@@ -114,6 +114,12 @@
 
 DATABASES = {"default": env.db_url()}
 
+# Caches
+CACHES = {"default": env.cache_url(default="locmemcache://")}
+SESSION_ENGINE = "django.contrib.sessions.backends.cached_db"
+DJANGO_REDIS_IGNORE_EXCEPTIONS = True
+DJANGO_REDIS_LOG_IGNORED_EXCEPTIONS = True
+
```
felixrindt, line 122: So what happens if I do a fresh checkout and try to start ephios? Does it fail anyway?

jeriox: if you set up a development environment, no CACHE_URL is set and therefore the default locmemcache is used. You have to actively set the CACHE_URL to point to your redis instance to enable it. If it cant reacht the cache, it will fail silently and just use the database.

felixrindt (approved):
We'll need to pay close attention that nothing breaks with the settings cache.

PR #365, opened 2021-03-11
add notification framework, enable push notifications

### LINKED ISSUE #236: Send push notifications
As a user, I expect the application to send notifications via the Notifications API (https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API) once I enable them. I expect to receive notifications for the same things that I receive emails for.
This could be implemented by https://github.com/safwanrahman/django-webpush

This could also be combined with #237 with https://pypi.org/project/django-pwa-webpush/

jeriox:
closes #236 
additional thoughts on the matter in the issue

current status: everything functional, but does not respect user preferences

TODO:

- [x] build settings and view for type/backend matrix
- [x] respect settings for sending out notifications
- [x] make more sturdy (maybe catch exceptions while sending and mark those notifications as failed)
- [x] build push backend
- [x] think about how to get the vapid keys into django
- [x] send notifications for confirmed participations to responsible when method does not use requested_state
- [x] write tests
- [ ] https://github.com/ephios-dev/ephios/pull/346#discussion_r592502443


ephios/core/signals.py
```diff
@@ -49,6 +48,16 @@
 If all forms are valid, `save` will be called on your form.
 """
 
+register_notification_types = Signal()
```
jeriox, line 51: @felixrindt can you help me here? I tried this with PluginSignal(), but as I'm accessing the signal already while migrating, the global preferences have not been set up at that point and everything breaks (have a look at https://github.com/ephios-dev/ephios/runs/2107604091). It works this way, do you think that's okay?

ephios/core/signals.py
```diff
@@ -69,4 +78,49 @@ def register_base_consequence_handlers(sender, **kwargs):
     dispatch_uid="ephios.core.signals.send_participation_state_changed_mail",
 )
 def send_participation_state_changed_mail(sender, instance, **kwargs):
-    mail.participation_state_changed(instance)
+    from ephios.core.models import AbstractParticipation
+
+    if instance.state == AbstractParticipation.States.CONFIRMED:
+        from ephios.core.notifications.types import ParticipationConfirmedNotification
+
+        ParticipationConfirmedNotification.send(instance)
```
jeriox, line 86: @felixrindt you'd prefer sending those from the disposition and not from the post_save, don't you?

ephios/core/signals.py
felixrindt: Where is this used in migrations? Can you avoid that?

ephios/extra/management/commands/send_notifications.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        for backend in all_notification_backends():
+            for notification in Notification.objects.filter(failed=False):
+                if backend.can_send(notification) and backend.user_prefers_sending(notification):
+                    try:
+                        backend.send(notification)
+                    except Exception as e:
```
jeriox, line 17: pylint doesn't allow this, but I don't know what to use here

ephios/core/signup/methods.py
```diff
@@ -359,9 +359,13 @@ def perform_signup(self, participant: AbstractParticipant, **kwargs) -> Abstract
...
         `kwargs` may contain further instructions from e.g. a form.
         """
+        from ephios.core.notifications.types import ResponsibleParticipationRequested
+
         if errors := self.get_signup_errors(participant):
             raise ParticipationError(errors)
-        return self.get_participation_for(participant)
+        participation = self.get_participation_for(participant)
+        ResponsibleParticipationRequested.send(participation)
+        return participation
```
felixrindt, line 368: Sending this notification fails with instant signup as the participation doesn't have an id here so looking it up while sending fails. Creating the notification object probably needs to happen somewhere else.
Lets include a regression test?!

felixrindt (changes requested):
I'd love to see a nicer `email_base.html`, with proper heading, footer, logo and design, but that's another issue I guess.

ephios/extra/management/commands/send_notifications.py
```diff
@@ -0,0 +1,24 @@
+import traceback
+
+from django.core.mail import mail_admins
+from django.core.management import BaseCommand
+
+from ephios.core.models.users import Notification
+from ephios.core.notifications.backends import all_notification_backends
+
+
+class Command(BaseCommand):
```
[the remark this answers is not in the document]
jeriox: definetly postpone...

ephios/extra/management/commands/send_notifications.py
jeriox: my thought was that a lot could happen until the notification gets send out, e.g. the event could have been deleted which would raise Event.DoesNotExist.
what would you suggest to handle this?

ephios/core/signup/methods.py
jeriox: can you explain how to reproduce this? for me notifications with instant_confirmation work flawlessly

ephios/extra/management/commands/send_notifications.py
felixrindt: It's probably good to catch all exceptions, but we should log them.

ephios/core/signup/methods.py
felixrindt: Showed you in the call

.env.example
```diff
@@ -9,4 +9,4 @@ SITE_URL=http://localhost:8000
 EMAIL_URL=dummymail://
 DEFAULT_FROM_EMAIL=webmaster@localhost
 SERVER_EMAIL=root@localhost
-ADMINS=Root <root@localhost>
+ADMINS=Root User <root@localhost.dev>
```
felixrindt, line 12: I should probably create an issue for the "broken" email regex in the library...

ephios/core/signup/methods.py
```diff
@@ -275,6 +275,8 @@ def check_conflicting_shifts(method, participant):
 
 
 class BaseSignupMethod:
+    # pylint: disable=too-many-public-methods
```
felixrindt, line 278: We could just reduce the public interface. No outside user needs to see the list of signup_checkers, only subclasses. Hence make the corresponding methods protected (prepend an underscore to the identifier)...

felixrindt (changes requested):

ephios/core/notifications/backends.py
```diff
@@ -0,0 +1,84 @@
...
+    for _, backends in register_notification_backends.send_to_all_plugins(None):
+        yield from (b() for b in backends)
+
+
+def enabled_notification_backends():
+    for _, backends in register_notification_backends.send(None):
+        yield from (b() for b in backends)
+
+
+class AbstractBackend:
```
felixrindt, line 20: Can we do 
```suggestion
class AbstractNotificationBackend:
```

and also expand the name for subclasses?

ephios/core/notifications/backends.py
```diff
@@ -0,0 +1,84 @@
...
+
+    @classmethod
+    def send(cls, notification):
+        notification_type = notification_type_from_slug(notification.slug)
+        email = EmailMultiAlternatives(
+            to=[cls._get_mailaddress(notification)],
+            subject=notification_type.get_subject(notification),
+            body=notification_type.as_plaintext(notification),
+        )
+        email.attach_alternative(notification_type.as_html(notification), "text/html")
```
felixrindt, line 69: I was asking myself, why `as_html` isn't a method on `Notification`?
Couldn't you just fetch the notification type and render the contents there instead of having to do this in the backend?
It's one more call, but it seems more clean to me.

ephios/core/notifications/types.py
```diff
@@ -0,0 +1,340 @@
...
+        for user in get_users_with_perms(event, only_with_perms_in=["view_event"]):
+            notifications.append(
+                Notification(slug=cls.slug, user=user, data=dict(event_id=event.id, **kwargs))
+            )
+        Notification.objects.bulk_create(notifications)
+
+    @classmethod
+    def get_subject(cls, notification):
+        event = Event.objects.get(pk=notification.data.get("event_id"))
+        return _("New {type}: {title}").format(type=event.type, title=event.title)
```
felixrindt, line 136: Translation here will be tricky, as discussed a few months ago.
Instead of "Neuer Dienst", "Neuer Ausbildung", "Neuer Biwak" we will have to e.g. drop the "new", prepend a "Rückmeldung erwünscht", or append an "added", like "Biwak hinuzgefügt: ...".

ephios/core/notifications/types.py
```diff
@@ -0,0 +1,340 @@
...
+        return _(
+            "You're receiving this email because a new account has been created for you at ephios.\n"
+            "Please go to the following page and choose a password: {url}\n"
+            "Your username is your email address: {email}\n"
+        ).format(url=urljoin(settings.SITE_URL, reset_link), email=notification.user.email)
+
+    @classmethod
+    def as_html(cls, notification):
+        return render_to_string(
+            "core/new_account_email.html",
```
felixrindt, line 112: Other mail templates are in the `mails` folder, right? Why isn't this one?

ephios/core/notifications/types.py
```diff
@@ -0,0 +1,340 @@
...
+        )
+        Notification.objects.create(
+            slug=cls.slug,
+            user=user,
+            data=dict(participation_id=participation.id, email=participation.participant.email),
+        )
+
+    @classmethod
+    def get_subject(cls, notification):
+        return _("Participation confirmed")
```
felixrindt, line 180: I think I'd like to see the event title in the subject.

ephios/core/notifications/types.py
```diff
@@ -0,0 +1,340 @@
...
+        )
+        Notification.objects.create(
+            slug=cls.slug,
+            user=user,
+            data=dict(participation_id=participation.id, email=participation.participant.email),
+        )
+
+    @classmethod
+    def get_subject(cls, notification):
+        return _("Participation rejected")
```
felixrindt, line 209: I think I'd like to see the event title in the subject.

ephios/core/notifications/types.py
```diff
@@ -0,0 +1,340 @@
...
+            )
+        Notification.objects.bulk_create(notifications)
+
+    @classmethod
+    def get_subject(cls, notification):
+        participation = AbstractParticipation.objects.get(
+            id=notification.data.get("participation_id")
+        )
+        participation_state = participation.get_state_display()
+        return _("Participation {state}").format(state=participation_state)
```
felixrindt, line 254: Again, I think I prefer more context in the subject, like participant name and event title. 
Using the state display here is a nice idea, lets hope it works in translation.

ephios/core/notifications/types.py
```diff
@@ -0,0 +1,340 @@
...
+            )
+        return _("{participant} signed up for {shift}").format(
+            participant=participation.participant, shift=participation.shift
+        )
+
+
+class EventReminderNotification(AbstractNotificationHandler):
+    slug = "ephios_event_reminder"
+    title = _("An event has vacant spots")
+    unsubscribe_allowed = False
```
felixrindt, line 277: Why can't I unsubscribe to this?

ephios/core/notifications/types.py
```diff
@@ -0,0 +1,340 @@
...
+        ).format(
+            title=event.title,
+            start=date_format(event.get_start_time(), "SHORT_DATETIME_FORMAT"),
+            end=date_format(event.get_end_time(), "SHORT_DATETIME_FORMAT"),
+            url=urljoin(settings.SITE_URL, event.get_absolute_url()),
+        )
+
+
+class CustomParticipantNotification(AbstractNotificationHandler):
+    slug = "ephios_confirmed_participants"
```
felixrindt, line 312: I think that slug can fit the class better.

felixrindt (changes requested):
Will look at the rest later :)

pyproject.toml
```diff
@@ -70,7 +71,7 @@ disable = """C0330, C0301,
 duplicate-code, attribute-defined-outside-init, missing-module-docstring, missing-class-docstring,
 no-member, invalid-name, import-outside-toplevel, unused-argument, too-many-ancestors, missing-function-docstring,
 too-few-public-methods, no-self-use, too-many-arguments, cyclic-import, inconsistent-return-statements,
-useless-object-inheritance
+useless-object-inheritance, logging-fstring-interpolation
```
felixrindt, line 74: Hmm I agree that we don't really benefit from deferred string formatting, which is the reason pylint wants us to avoid fstrings in logging, right?

.env.example
felixrindt: https://github.com/web-push-libs/vapid/pull/90

felixrindt (approved):

Issue #280, opened 2021-01-21
Are you sure?-Prompts


jeriox:
As a planner, I'd like to be asked if I am sure I want to discard my changes when trying to leave a page with a changed form without saving.

garinm90:
Hey @jeriox I can add this feature if you'd like.

jeriox:
Hey @garinm90, cool that you want to help! We are still in an early stage and haven't received contributions from outside the team yet, so things are sometimes a bit messy and not well documented. Make sure to check out the [contribution page](https://docs.ephios.de/en/latest/development/contributing.html) for information on how to set up the project. Feel free to contact us if you have any questions, you can find our mail addresses in the pyproject.toml

PR #366, opened 2021-03-11
280 areyousure

garinm90:
Are you sure?-Prompts #280

Add are you sure jQuery plugin.
Load this plugin on form pages. Add classes to inputs that dynamically create new forms to be monitored.

### DIFF
--- a/ephios/static/ephios/js/areyousure_form.js
+++ b/ephios/static/ephios/js/areyousure_form.js
@@ -1,3 +1,7 @@
 $(document).ready(function () {
     $('form').areYouSure();
+    // Listener for inputs that can dynamically add new forms and trigger a scan of the new form
+    $('.add-form').click(function() {
+        $('form').trigger('reinitialize.areYouSure');
+      });
 });
--- a/ephios/static/ephios/js/areyousure_form.js
+++ b/ephios/static/ephios/js/areyousure_form.js
@@ -1,0 +1,3 @@
+$(document).ready(function () {
+    $('form').areYouSure();
+});
--- a/ephios/core/templates/core/event_form.html
+++ b/ephios/core/templates/core/event_form.html
@@ -64,3 +64,8 @@
         </div>
     </form>
 {% endblock %}
+
+{% block javascript %}
+    <script type="text/javascript" src="{% static "areyousure/are-you-sure.js" %}"></script>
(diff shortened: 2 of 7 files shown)


felixrindt (commented):
Welcome to the project :)

ephios/static/ephios/js/areyousure_form.js
```diff
@@ -0,0 +1,7 @@
+$(document).ready(function () {
+    $('form').areYouSure();
+    // Listener for inputs that can dynamically add new forms and trigger a scan of the new form
+    $('.add-form').click(function() {
+        $('form').trigger('reinitialize.areYouSure');
+      });
+});
```
felixrindt, line 7: Couldn't we just look for forms (or their submit buttons) in `main.js` and apply areyousure there? I saw this here:

https://github.com/pretix/pretix/blob/a897f60fc5a22d9eb9bf4107111eb8ea05d09b3b/src/pretix/static/pretixcontrol/js/ui/main.js#L874

That way, we wouldn't need to (remember to) include areyousure in every form based view. Also, if we get false positives, we could add a class to exclude those particular elements. What do you think?

garinm90: Sure thing! I'll update and push those changes.

ephios/static/ephios/js/main.js
```diff
@@ -9,6 +9,7 @@ function handleForms(elem) {
     }).on("formAdded", "div", function (event) {
         handleForms($(event.target));
     });
+    $('form').trigger('reinitialize.areYouSure');
```
felixrindt, line 12: I moved this line up here, as this function is used for all dynamically added inputs and it now also fully works in the shift edit view when changing the signup method (which was done using a select and some view-specific JS).

felixrindt (approved):
@garinm90 Nice! Using the plugin is was very simple! Only thing on my mind now is internationalization. How do we get an e.g. german translation into the alert?

garinm90:
> @garinm90 Nice! Using the plugin is was very simple! Only thing on my mind now is internationalization. How do we get an e.g. german translation into the alert?

After doing some digging looks like Django offers a solution for this.
https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#internationalization-in-javascript-code

I'll look into adding this so future JavaScript modules will also have access to these translations!

felixrindt:
By the way, we are already using jsi18n...

PR #369, opened 2021-03-14
Fix can't sign up for back-to-back shifts

### LINKED ISSUE #367: Shifts back to back are marked as conflicting
As a user, I want to be able to sign up to two shifts that are back to back (e.g. one shift ends at 12:00, the other starts at 12:00). Currently, these are marked as conflicting, so I am not able to sign up.

felixrindt:
closes #367


jeriox (approved):
a test for this edge case would be nice ;)

PR #371, opened 2021-03-15
bypass Plugin Signal for signup methods

felixrindt:


ephios/core/views/shift.py
```diff
@@ -40,10 +41,31 @@ def post(self, *args, **kwargs):
             from ephios.core.signup import signup_method_from_slug
 
             signup_method = signup_method_from_slug(self.request.POST["signup_method_slug"])
-            configuration_form = signup_method.get_configuration_form(self.request.POST)
+        except KeyError as e:
```
felixrindt, line 44: I think this `post` method is too complex now.

jeriox: I think it's still okay

jeriox (approved):

PR #385, opened 2021-03-21
restrict settings view to general settings

jeriox:


felixrindt (approved):
How did this come up actually?

PR #391, opened 2021-03-22
Add auto qualification plugin

### LINKED ISSUE #263: Select qualifications to be granted from event
As a planner I expect to select the qualifications that users should acquire after participating in an event if the event type allows so.

### LINKED ISSUE #331: Add padding to "Add ..." buttons to Users and Groups pages
It looks horrible right now. I'm in physical pain. Let's fix it. (It's already correct on the Events page btw.)

felixrindt:
closes #263

Oh and also closes #331 because it was really annoying me too.


ephios/plugins/eventautoqualification/consequences.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    Creates
+    """
+    relevant_participations = (
+        AbstractParticipation.objects.filter(
+            state=AbstractParticipation.States.CONFIRMED,
+            shift__end_time__lt=timezone.now(),
+            shift__event__auto_qualification_config__isnull=False,
+        )
+        .filter(~Q(shift__event__auto_qualification_config__handled_for=F("id")))
```
felixrindt, line 27: Use exclude

jeriox:
grafik

ephios/plugins/eventautoqualification/consequences.py
```diff
@@ -0,0 +1,85 @@
...
+        .select_related("shift", "shift__event", "shift__event__auto_qualification_config")
+    )
```
felixrindt, line 29: We figured it would be smart to offer a `participations_finished` signal from core.

ephios/core/utils.py
```diff
@@ -1,9 +1,17 @@
+import logging
```
jeriox, line 1: this file is a bit weird IMO. why do we have a utils.py when the whole extra module basically provides utils?

ephios/core/utils.py
```diff
@@ -19,3 +27,26 @@ def password_changed(self, password, user):
...
+    """
+    from ephios.core.signals import participation_finished
+
+    relevant_participations = AbstractParticipation.objects.filter(
+        state=AbstractParticipation.States.CONFIRMED,
+        shift__end_time__lt=timezone.now(),
+        finished=False,
+    ).select_related("shift", "shift__event")
+    for participation in relevant_participations:
+        try:
```
jeriox, line 44: this means that it isn't guaranteed that the signal is only called once for the participation, is it?
e.g. the first receiver sends out some notifications, the second receiver raises. The participation will not be marked as finished and the signal will be called again for it the next time

ephios/plugins/eventautoqualification/consequences.py
```diff
@@ -0,0 +1,69 @@
...
+        return
+
+    # check if the required shift participations exist
+    requirements_met = False
+    if mode == EventAutoQualificationConfiguration.Modes.ANY_SHIFT:
+        requirements_met = True
+    elif mode == EventAutoQualificationConfiguration.Modes.LAST_SHIFT:
+        requirements_met = (
+            participation.shift
+            == sorted(event.shifts.all(), key=operator.attrgetter("end_time"))[-1]
```
jeriox, line 28: why sort in python and not on the queryset?

ephios/plugins/eventautoqualification/consequences.py
```diff
@@ -0,0 +1,69 @@
...
+        # count participant hashes in finished participations (and this one)
+        participant_hash_counter = Counter(
+            hash(participation.participant)
+            for participation in AbstractParticipation.objects.filter(
+                Q(finished=True) | Q(id=participation.id)
+            )
+        )
+        # if the amount of this participant in confirmed past shifts is equal to the the count of all shifts
+        # then this participant attended every shift
+        requirements_met = participant_hash_counter[hash(participation.participant)] == len(
```
jeriox, line 40: why use len() and not shifts.count()?

ephios/plugins/eventautoqualification/signals.py
```diff
@@ -0,0 +1,22 @@
...
+from ephios.plugins.eventautoqualification.consequences import create_qualification_consequence
+from ephios.plugins.eventautoqualification.forms import EventAutoQualificationForm
+
+
+@receiver(
+    event_forms,
+    dispatch_uid="ephios.plugins.eventautoqualification.signals.configuration_form",
+)
+def configuration_form(sender, event, request, **kwargs):
+    if not request.user.has_perm("core.change_userprofile"):
```
jeriox, line 13: maybe it would be nice to display a hint like "this event will grant xy, but you don't have the permission to manage qualification grants"?
but that's nice to have, you can also move that to an issue

jeriox (commented):
i love it <3

ephios/core/models/events.py
```diff
@@ -123,7 +123,13 @@ def labels_dict(cls):
...
     )
     state = IntegerField(_("state"), choices=States.choices)
-    data = models.JSONField(default=dict)
+    data = models.JSONField(default=dict, verbose_name=_("Signup data"))
+
+    """
+    The finished flag is used to make sure the participation_finished signal is only sent out once, even
+    if the shift time is changed afterwards.
+    """
+    finished = models.BooleanField(default=False)
```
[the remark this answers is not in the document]
felixrindt: But the participation is, right? It's a strange case, like how would you handle that sensibly?

ephios/plugins/eventautoqualification/consequences.py
felixrindt: To not cause another dB hit

ephios/plugins/eventautoqualification/consequences.py
felixrindt: To not cause another DB hit

ephios/core/models/events.py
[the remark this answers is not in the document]
jeriox: the handling is fine, i was just thinking about the term "finished" which is technically not correct in this case. But as it is very descriptive I'm fine with that

ephios/core/__init__.py
```diff
@@ -1,7 +1,12 @@
 from django.apps import AppConfig
-from dynamic_preferences.registries import preference_models
+from dynamic_preferences.registries import PerInstancePreferenceRegistry, preference_models
 
-from .registries import event_type_preference_registry
+
+class EventTypeRegistry(PerInstancePreferenceRegistry):
```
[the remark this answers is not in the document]
felixrindt: I thought its not that important to have it's own file. We could also move it to the model...
So it's just a refactoring, no necessity.

ephios/core/signals.py
```diff
@@ -86,20 +93,30 @@ def register_base_consequence_handlers(sender, **kwargs):
...
     return CORE_NOTIFICATION_BACKENDS
 
 
-@receiver(periodic_signal)
+@receiver(periodic_signal, dispatch_uid="ephios.core.signals.send_notifications")
 def send_notifications(sender, **kwargs):
     send_all_notifications()
+
+
+periodic_signal.connect(
```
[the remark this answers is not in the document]
felixrindt: Well I didn't want to cludder the file, yes.

ephios/core/utils.py
felixrindt: I think there is a difference between library-like functions, like custom widgets and mixins, and actual business logic. This file  - in `ephios.core` - has business logic, so I don't think it's "extra". 

Both artifacts in this module hook into one thing happening and trigger another thing. We could rename it to `services.py`, or, as we should probably avoid growing a random assortment of things in one file, could create an `ephios.core.services` package containing this sort of stuff. That package might also house the `notifications` package, and some of the async setup later on.

ephios/core/utils.py
felixrindt: That's correct, though I first intended to do it differently. 
Currently, everything is wrapped in a transaction, so that should roll back changes of the first receiver, depending on what it did.
There's definitely a discussion here.

* Single try, no transaction: good for stuff that calls external APIs or something similar that won't be rolled back with a transaction. Though I think these calls should be idempotent anyway.
* Multiple tries, transaction rollback: Somewhat better for stuff like that works with the DB, like models (notifications), although they could do their own handling, and they would probably fail again?!

 I'm not sure, I would be fine with keeping it as is. We could also use the signal's `sender` argument to let receivers decide:
 
 ```python
 try:
    with transaction.atomic():
        participation_finished.send(TRANSACTION, participation=participation)
        participation.finished = True
        participation.save()
except Exception:
    ...
else:
    participation_finished.send(SINGLE_CALL, participation=participation)
 ```
 
 What do you think?

ephios/core/__init__.py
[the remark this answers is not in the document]
jeriox: I just wasn't sure if it is relevant regarding the moment of execution

ephios/core/utils.py
jeriox: I think I'm also fine with it for the moment

ephios/plugins/eventautoqualification/signals.py
felixrindt: I built that.

ephios/core/utils.py
felixrindt: I will add a notice in the signal's docstring.

ephios/core/__init__.py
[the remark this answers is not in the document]
felixrindt: I figured everything still works.

ephios/core/utils.py
felixrindt: I did decide to switch to the single call.

jeriox (approved):

PR #393, opened 2021-03-23
add signal for event bulk actions

jeriox:


ephios/core/templatetags/event_extras.py
```diff
@@ -77,3 +80,14 @@ def event_signup_state_counts(event, user):
                 counter[participation.state] += 1
                 break
     return counter
+
+
+@register.simple_tag(name="event_bulk_actions")
+def event_bulk_actions():
+    return format_html(
+        reduce(
+            lambda result, item: result + "".join(item[1]),
```
felixrindt, line 89: Can you use nested tuple unpacking? 
```suggestion
            lambda html, (receiver, buttons): html + "".join(buttons),
```

felixrindt (approved):

ephios/core/templatetags/event_extras.py
jeriox: tuple parameter unpacking is not supported in python3

felixrindt: Oh okay. Do you see another way to make that more readable?

jeriox: is this comment based on the new code?

felixrindt: Probably not

felixrindt (approved):

ephios/core/signals.py
```diff
@@ -64,6 +64,14 @@
 This signal is called periodically, at least every 15 minutes.
 """
 
+register_event_bulk_action = PluginSignal()
+"""
+This signal is sent out to get a list of actions that a user can perform on a list of events.
+Receivers should return a list of actions. Each action is represented by a dict with the keys ``url``, ``title`` and ``icon``.
```
felixrindt, line 70: `title` should be `label` maybe?

PR #395, opened 2021-03-23
add color to eventtype

### LINKED ISSUE #336: Colors for event types
As a manager 
I want to assign colors to event types
so that they can be better differentiated visually in lists and filters.

jeriox:
closes #336


ephios/core/templatetags/event_extras.py
```diff
@@ -77,3 +78,28 @@ def event_signup_state_counts(event, user):
                 counter[participation.state] += 1
                 break
     return counter
+
+
+@register.simple_tag(name="eventtype_colors")
+def eventtype_colors(request):
+    html = f"<style nonce='{request.csp_nonce}'>"
+    for eventtype in EventType.objects.all():
```
felixrindt, line 86: Is this a query we can avoid by caching the results of this function?
This could be as simple as a time based cache for the template html...

felixrindt (commented):
The event_detail page could also use the color badge IMO.

ephios/core/templatetags/event_extras.py
jeriox: tell me moer about time based cache

felixrindt: *in :robot: voice*: Ok, here is more about time based caches: https://docs.djangoproject.com/en/3.1/topics/cache/#template-fragment-caching

felixrindt:
> The event_detail page could also use the color badge IMO.

what about this?

jeriox:
> 
> 
> > The event_detail page could also use the color badge IMO.
> 
> what about this?

oh I missed this, just looked at the review comments

felixrindt (approved):

PR #396, opened 2021-03-25
fix metadata error for python3.8

jeriox:
importlib.metadata is broken on python3.8, see https://github.com/apache/airflow/issues/12855 for explanation


felixrindt (approved):

ephios/settings.py
```diff
@@ -1,7 +1,11 @@
 import copy
 import os
 from email.utils import getaddresses
-from importlib import metadata
+
+try:
+    import importlib_metadata  # importlib is broken on python3.8, using backport
```
felixrindt, line 6: :roll_eyes:

PR #402, opened 2021-03-27
fix error emails that are not being sent

jeriox:


ephios/settings.py
```diff
@@ -188,6 +188,25 @@
 SERVER_EMAIL = env.str("SERVER_EMAIL")
 ADMINS = getaddresses([env("ADMINS")])
 
+# logging
```
felixrindt, line 191: wow

felixrindt (approved):

PR #403, opened 2021-03-27
fix template cache

jeriox:


ephios/core/forms/events.py
```diff
@@ -201,6 +203,11 @@ class Meta:
         fields = ["title", "can_grant_qualification", "color"]
         widgets = {"color": ColorInput()}
 
+    def save(self, commit=True):
+        cache_key = make_template_fragment_key("eventtype_colors")
+        cache.delete(cache_key)
+        return super().save(commit)
```
felixrindt, line 209: Why not use a post_save signal? To me, the place where the css is generated seems like a more fitting place for this to live.

jeriox: couldn't get this to work, the cache is not invalidated with the same code that works here.

PR #404, opened 2021-03-27
Model Change Logging

felixrindt:
* [x] tests
* [x] Full-text-search in the FilterForm
* [x] rename `db_instance` to be a copy of the instance from the `post_init` signal
* [x] views
  * [x] generic view
  * [x] teaser on index page
* [x] refactor json structure `{str: ..., fields: [ {slug: ..., data: ...}, ... ]}`
* [x] refactor |linkify for get_absolute_url
* [ ] refactor object select
  * [ ] heterolog heavy select2
* [x] `view_logentry` permission for managers? 
* [x] log group permissions

Ideas: 
* clickable links to absolute urls -> doesn't really fit our needs
* diffs on large text changes -> difflib exists, but it's more work than I want to currently spend for the benefit.
* clicking an object link makes it the filter?! --> not useful enough


jeriox:
grafik
the message may not be very intuitive

ephios/modellogging/models.py
```diff
@@ -0,0 +1,252 @@
...
+    M2MLogRecorder,
+    ModelFieldLogRecorder,
+    capitalize_first,
+    register_log_recorders,
+)
+
+# pylint: disable=protected-access
+
+
+class LogJSONEncoder(JSONEncoder):
```
jeriox, line 27: isn't that what we are already doing in our CustomJSONEncoder?

ephios/modellogging/templates/modellogging/logentries.html
```diff
@@ -0,0 +1,32 @@
...
+<ul class="list-group">
+    {% for log_group in logged_object.grouped_logentries %}
+        <li class="list-group-item">
+            <p>
+                <span class="pr-3">
+                    <span class="far fa-clock" aria-hidden="true"></span>
+                    {{ log_group.0.datetime|date:"SHORT_DATETIME_FORMAT" }}
+                </span>
+                {% if log_group.0.user %}
+                    {% if log_group.0.user.is_manager %}
```
jeriox, line 13: we don't have an is_manager flag

jeriox (commented):
lgtm

ephios/modellogging/models.py
felixrindt: Not exactly. I just changed the JSONEncoders again. 
Also, as this app is designed in a standalone way, I don't want to import ephios code.

felixrindt: (although I am using the one permission function...)

felixrindt:
We now just need to decide where to show the log and to whom.

jeriox:
> 
> 
> We need to decide where to show the log and to whom.
> Rendering the log requires - due to the architecture, a lot of queries.
> 
>     * Events, Shifts and Participations are currently logged together. I propose a separate view visible to event responsibles where they can view the log.

sounds good

>     * We can add logging to users (but not groups, sadly, as we can't access that model) as well, though I wouldn't show the log anywhere specifically. Instead
> 
>     * We can create a View where you can filter logs based on content type, timeframe and acting user. Useful for support staff and managers(?)/staff users!

also sounds good. I still like the idea of having a few of the latest actions on the homepage for managers with a link to the view that you just mentioned.

ephios/settings.py
```diff
@@ -189,22 +191,25 @@
...
 
 # logging
-LOGGING = {
-    "version": 1,
-    "disable_existing_loggers": True,
-    "handlers": {"mail_admins": {"level": "ERROR", "class": "django.utils.log.AdminEmailHandler"}},
-    "loggers": {
-        "django": {
+if not DEBUG:
+    LOGGING = {
```
felixrindt, line 195: Oh I need to update the diff here.

jeriox:
After clicking on "view edit history" on an event with an empty history, this happens
grafik

jeriox:
after deleting a shift, this happens
grafik

jeriox:
grafik
Not sure if we should log that? This results in an exact history on when each user accessed ephios

ephios/core/views/log.py
```diff
@@ -0,0 +1,88 @@
...
+    object_type = forms.ChoiceField(
+        choices=[("", "---------")],
+        required=False,
+        label=_("Concerns"),
+        widget=Select2Widget,
+    )
+    object_id = forms.IntegerField(required=False, widget=forms.HiddenInput)
+    user = forms.ModelChoiceField(
+        queryset=UserProfile.objects.all(),
+        widget=Select2Widget,
```
jeriox, line 26: we are using heavy select2 widgets for users everywhere else

ephios/settings.py
```diff
@@ -189,22 +191,25 @@
 ADMINS = getaddresses([env("ADMINS")])
 
 # logging
-LOGGING = {
-    "version": 1,
-    "disable_existing_loggers": True,
-    "handlers": {"mail_admins": {"level": "ERROR", "class": "django.utils.log.AdminEmailHandler"}},
-    "loggers": {
-        "django": {
+if not DEBUG:
```
felixrindt, line 194: So I was annoyed that I didn't get a stacktrace, but a console email for errors.

jeriox:
grafik

jeriox:
grafik
string representation could be more descriptive

jeriox:
grafik
don't really like the string representation

ephios/core/views/event.py
```diff
@@ -294,6 +295,14 @@ def post(self, *args, **kwargs):
 
 class HomeView(LoginRequiredMixin, TemplateView):
     template_name = "core/home.html"
+    LOGENTRY_COUNT = 5
+
+    def get_context_data(self, **kwargs):
+        if self.request.user.is_staff:
+            logentries = LogEntry.objects.all()[: self.LOGENTRY_COUNT]
```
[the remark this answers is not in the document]
felixrindt: it uses default ordering, yes.

ephios/core/views/event.py
```diff
@@ -294,6 +295,14 @@ def post(self, *args, **kwargs):
 
 class HomeView(LoginRequiredMixin, TemplateView):
     template_name = "core/home.html"
+    LOGENTRY_COUNT = 5
+
+    def get_context_data(self, **kwargs):
```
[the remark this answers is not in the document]
felixrindt: You mean the queryset, not the html itself?

PR #405, opened 2021-03-27
add calendar view for events

### LINKED ISSUE #241: Display events as calendar view
As a user, I expect the system to have a view where i can browse a calendar view with the events. There should be some information rendered for each event and I expect to be redirected to the event detail page on click.

As a manager I expect to have a setting where I can choose if the calendar or the list view should be the default.

jeriox:
closes #241


ephios/static/ephios/css/main.css
```diff
@@ -102,3 +102,15 @@ This hack fixes that. Inspired by https://github.com/select2/select2/issues/4220
...
+    table-layout: fixed;
+}
+
+.calendar-row-height {
+    height: 100px;
+}
+
+.today {
+    background-color: #cbcbcb;
+}
```
felixrindt, line 116: This should probably be in a calendar-specific css file.

ephios/core/views/event.py
```diff
@@ -316,3 +320,24 @@ def form_valid(self, form):
...
         return redirect(self.object.get_absolute_url())
+
+
+class EventCalendarView(TemplateView):
+    template_name = "core/event_calendar.html"
+
+    def get_context_data(self, **kwargs):
+        year, month = kwargs["year"], kwargs["month"]
+        shifts = Shift.objects.filter(start_time__month=month, start_time__year=year)
+        calendar = ShiftCalendar(shifts, locale=get_language())
```
felixrindt, line 331: On my machine, there is not locale `de` installed, just `de_DE`, so it errored away at first.

jeriox: ahhh i hate languages.
We are using `de` everywhere because this is broader than `de_DE` and your machine should choose `de_DE` anyway. We are also only compiling translations for `de`.

ephios/static/ephios/css/main.css
felixrindt: I think `.today` does not make clear that it is used for a calendar and it might clash with other css in the future.
 Either specify the selectors or move to a specific css file.

ephios/core/views/event.py
felixrindt: Well we should have a test for it then.

felixrindt (changes requested):

ephios/core/views/event.py
jeriox: test for what exactly?

felixrindt: The view

ephios/core/calendar.py
```diff
@@ -0,0 +1,36 @@
+from calendar import LocaleHTMLCalendar
+from datetime import date
+from itertools import groupby
+
+
+class ShiftCalendar(LocaleHTMLCalendar):
```
felixrindt, line 6: Lets use `HTMLCalendar` as superclass and implement the formatting methods using django tools.

felixrindt (changes requested):

felixrindt (approved):

PR #412, opened 2021-03-31
do not display past shifts on homepage

jeriox:


ephios/core/templatetags/user_extras.py
```diff
@@ -52,5 +54,9 @@ def participant_conflicting_shifts(participant, shift):
 def shifts_needing_disposition(user):
     return Shift.objects.filter(
         participations__state=AbstractParticipation.States.REQUESTED,
-        event__in=get_objects_for_user(user, perms=["core.change_event"]),
+        event__in=get_objects_for_user(
+            user,
+            perms=["core.change_event"],
+        ),
+        end_time__gt=datetime.now(),
```
felixrindt, line 61: Always use `django.utils.timezone.now` to get a timezone-aware datetime.
Maybe you also find other places where we did that incorrectly?!

felixrindt (changes requested):

ephios/core/templatetags/user_extras.py
jeriox: found one other place.
It's also widely used in the event copy stuff, but I don't want to touch that rn

felixrindt: Well there it was intentional wasn't it?

felixrindt (approved):

PR #413, opened 2021-03-31
add changelog

jeriox:
release-drafter loads the template file from the default branch (aka main). This is why the check is currently failing. It _should_ work once this is merged, but there is no option to test it before.


felixrindt (approved):
Ok I'm excited to see what this does.

PR #428, opened 2021-04-12
fix ordering for user profiles and qualification grants

jeriox:


ephios/plugins/basesignup/signup/common.py
```diff
@@ -114,7 +114,9 @@ def get_participation_display(self):
                 ", ".join(
                     participant.qualifications.filter(
                         category__in=relevant_qualification_categories
-                    ).values_list("title", flat=True)
+                    )
+                    .order_by("title")
```
felixrindt, line 118: Did you try ordering by Category? Might be sensible as well...

jeriox: good point

felixrindt (approved):

ephios/plugins/basesignup/signup/common.py
felixrindt: Now we can't change the ordering in the relevent-qualification settings, sadly.

PR #451, opened 2021-04-28
fix custom notification for deleted event

jeriox:


ephios/core/services/notifications/types.py
```diff
@@ -333,15 +333,21 @@ def send(cls, event: Event, content: str):
                 Notification(
                     slug=cls.slug,
                     user=user,
-                    data=dict(email=participant.email, event_id=event.id, content=content),
+                    data=dict(
+                        email=participant.email,
+                        event_id=event.id,
```
felixrindt, line 338: Do we need the id, still?

ephios/core/services/notifications/types.py
```diff
@@ -333,15 +333,21 @@ def send(cls, event: Event, content: str):
...
+                        content=content,
+                        event_title=event.title,
```
felixrindt, line 340: I must admit this feels like fixing symptoms. There are more occasions in which we will try to fetch objects in Notification sending that have been deleted in the meantime...

ephios/core/services/notifications/types.py
jeriox: not at the moment, but I think we should keep the ability to fetch the event from the db for sending to query for more information (if it exists)

ephios/core/services/notifications/types.py
jeriox: well usually you won't confirm people directly before deleting the event, so this will be the case where it happens IMO.
What would be your suggestion?

felixrindt: I think the best solution would be to send Notifications without delay, but that's for another day.

felixrindt (approved):

PR #471, opened 2021-05-07
add placeholder participations

### LINKED ISSUE #303: Placeholder participations
As a planner on the disposition view I expect to be able to add participations for people that don't have a user account (e.g. trainee, people from other organizations etc.).
I expect the view to have a field similar to the "add other"-box where I can enter the name of the guest participant. A PlaceholderParticipation should be created for this participant.

jeriox:
closes #303 

TODO

- [x] saving the disposition form violates a foreign key constraint when having a local user participation and a foreign participation both in the GETTING_DISPATCHED STATE
- [x] select2 field does not get cleared correctly anymore


felixrindt:
Could we do something like searching for an existing name and if nothing is found, create the external user with the search string to avoid creating external users that have an account?!

jeriox:
> 
> 
> Could we do something like searching for an existing name and if nothing is found, create the external user with the search string to avoid creating external users that have an account (or some similar flow)?!

yeah also thought about that. select2 supports that in theory, what do you think about the flow?

felixrindt:
> > 
> > 
> > Could we do something like searching for an existing name and if nothing is found, create the external user with the search string to avoid creating external users that have an account (or some similar flow)?!
> 
> yeah also thought about that. select2 supports that in theory, what do you think about the flow?
> https://select2.org/tagging

Hmm yeah I'd like to see it in action. Esp how it feels and behaves when searching existing users.

ephios/static/plugins/basesignup/js/disposition.js
```diff
@@ -1,4 +1,19 @@
...
+        createTag: function (params) {
+            console.log(params)
+            return {
+                id: params.term,
+                text: params.term,
+                guestUser: true
+            }
+        }
+    });
+
```
felixrindt, line 16: options needs bootstrap5 theme

ephios/core/signup/methods.py
```diff
@@ -126,6 +127,37 @@ def reverse_event_detail(self, event):
...
+            return None
+
+    def all_participations(self):
+        return AbstractParticipation.objects.none()
+
+    def reverse_signup_action(self, shift):
+        return reverse("core:signup_action", kwargs=dict(pk=shift.pk))
+
+    def reverse_event_detail(self, event):
+        return event.get_absolute_url()
```
felixrindt, line 152: Do these need to be implemented? And if, shouldn't it return something that makes it obvious that it's not a meaningful thing to do?

ephios/core/signup/methods.py
```diff
@@ -126,6 +127,37 @@ def reverse_event_detail(self, event):
         return event.get_absolute_url()
 
 
+@dataclasses.dataclass(frozen=True)
+class ForeignParticipant(AbstractParticipant):
```
felixrindt, line 131: Name could be:

* Placeholder
* Mock
* Strawman
* Oneoff
* Unidentified
* Anonoumous

I think Placeholder and unidentified resemble it best. 
The participation is there to stand in for a person without an actual account/method of identification or correlation.

felixrindt (commented):
I like the scope of it. 
UX wise it needs a thing indicating that a placeholder will be created.

ephios/static/plugins/basesignup/js/disposition.js
```diff
@@ -1,14 +1,17 @@
...
+            if (data.length === 0) {
+                data.push(tag);
+            }
         },
         createTag: function (params) {
-            console.log(params)
             return {
                 id: params.term,
-                text: params.term,
+                text: "Create \"" + params.term + "\" as placeholder",
```
felixrindt, line 13: this is not translated, right?

felixrindt (changes requested):

felixrindt (commented):
I get a `FOREIGN KEY constraint failed` when trying to save a disposition containing a placeholder participation. Do you get that too? There should be a test.

ephios/static/plugins/basesignup/js/disposition.js
```diff
@@ -1,4 +1,22 @@
 $(document).ready(function () {
+    $("#id_user").djangoSelect2({
+        theme: "bootstrap5",
+        insertTag: function (data, tag) {
+            // Insert the tag at the end of the results
+            if (data.length === 0) {
+                data.push(tag);
+            }
+        },
```
felixrindt, line 9: What does this do actually?

jeriox: the length check only shows the placeholder if there is no user matching the search string.
the data.push() adds the tag to the result list displayed to the user

felixrindt (approved):

PR #494, opened 2021-06-13
Design rework: bootstrap 5, scss, colors

### LINKED ISSUE #394: Switch to bootstrap 5
As of #334 we are using crispy-forms, which should make the change somewhat easier.

Required steps:

- [x] include the new source files and exclude the old ones
- [x] change the crispy template pack to bootstrap5 (https://github.com/django-crispy-forms/crispy-bootstrap5)
- [x] change forms where we manually build the structure (e.g. qualification grants)
- [x] change bootstrap classes that have been deprecated (see https://getbootstrap.com/docs/5.0/migration/ and the existing bootstrap5 branch in this repo which is wildly outdated)
- [x] fix JS stuff (especially tooltips)
- [x] include select2 bootstrap5 theme
- [x] think about whether we still need jQuery as bootstrap does not use it anymore

felixrindt:
closes #394

* [x] finish #394 checklist
* [x] message close button
* [x] tooltips
* [x] dropdowns
* [x] collapsables
* [x] restore use of crispy layout helper in forms
    * [x] userprofile
    * [x] group
    * [x] event notification (fix ugly js)
    * [x] working hours request


ephios/core/consequences.py
```diff
@@ -88,7 +88,7 @@ class WorkingHoursConsequenceHandler(BaseConsequenceHandler):
     def create(
         cls,
         user: UserProfile,
-        when: datetime,
+        when: date,
```
jeriox, line 91: is this breaking existing consequences?

ephios/static/ephios/scss/main.scss
```diff
@@ -0,0 +1,52 @@
...
+
+// Helpers
+@import "../../bootstrap/scss/helpers";
+
+// Utilities
+@import "../../bootstrap/scss/utilities/api";
+// scss-docs-end import-stack
+
+// 4. Add additional custom code here
+@import "random";
```
jeriox, line 52: is this some sort of convention? I don't like the term "random" here, would prefer "custom" or smth like that

ephios/core/consequences.py
felixrindt: this is just a change in type annotation. types are not enforced at runtime, and from what I saw, we are already only passing dates. Also, on the model, it's a date field.

ephios/static/ephios/scss/main.scss
felixrindt: Yeah I agree.

jeriox (approved):

PR #513, opened 2021-07-17
Design update

### LINKED ISSUE #57: Discuss signup method explanations to user
do we want to explain the active signup method to the user on the event detail page? it may be confusing when different things happen after hitting the register button without explanation because different signup methods do different things. could also be information overload.

We should also change the terminology for the event detail page:
Anmelden -> Teilnehmen

### LINKED ISSUE #397: Notification dot for event with requested participation
It would be great to have an indicator that looks a bit better.

### LINKED ISSUE #419: Display required qualifications in SectionBasedSignup to user
As a user viewing a shift with SectionBasedSignup I want to know which qualifications are required to participate, similar to how this is displayed for RequestConfirm. Optionallly I want to see the individual qualifications that are required for each section.

felixrindt:
* [x] ephios mannequin icons
  * [x] get figma colors in scss
  * [x] get the files
* [x] Find out how to show required qualification
  * [x] showing needed qualification in placeholder inline participation
  * [x] putting an exclamation icon at the start
  * [x] -> don't show the qualification as a label, but as detailed info for a placeholder spot. -> only works with min_count
* [x] render missing/free indicator
  * [x] section-based
  * [x] min-max-common
* [x] check how emails look 
* [x] try to increase test coverage
  * [x] section based PDF export
* [x] replace `Needed: None` with better indicator
* [x] check amount of queries on event list and detail
* [x] think about #419 
* [x] think about #397 
  * [x] the dot could be a small badge showing the amount of requests...
* [x] think about #57 

I advocate to close #397 as there is an indicator now that works IMO. We can open a new issue with a concrete idea.

The current solution also closes #419. I also propose to close #57. I renamed the instant button to say "Participate", but don't think an explanation should be required for the participant.


ephios/core/signup/methods.py
```diff
@@ -534,22 +536,41 @@ def render_configuration_form(self, *args, form=None, **kwargs):
 @dataclasses.dataclass()
 class SignupStats:
     requested_count: int
-    signed_up_count: int
-    missing: Optional[int]
-    free: Optional[int]
+    confirmed_count: int
+    missing: int
```
jeriox, line 540: why isn't missing optional anymore? if I don't have a minimun, I don't have a missing count

ephios/plugins/basesignup/templates/basesignup/section_based/fragment_state.html
```diff
@@ -2,13 +2,16 @@
...
+        </strong> {{ section.title }}</span>
+
         <div class="mb-3">
             {% for participation in section.participations %}
                 {% include "basesignup/participation_card_inline.html" %}
             {% endfor %}
-            {% for i in section.missing %}
-                {% if section.qualifications_label %}
+            {% for i in section.placeholder %}
+                {% if False and section.qualifications_label %}
```
jeriox, line 14: ???

ephios/core/signup/methods.py
felixrindt: So zero people are missing?!

jeriox: yeah just thought about the label in the event list, but we check for > 0 there anyways.

ephios/plugins/basesignup/templates/basesignup/section_based/fragment_state.html
felixrindt: This was me trying out how it looks without. Will remove...

felixrindt:
I vow to close #397 as there is an indicator now that works IMO. We can open a new issue with a concrete idea.

The current solution also closes #419. I also propose to close #57. I renamed the instant button to say "Participate", but don't think an explanation should be required for the participant.

ephios/plugins/basesignup/signup/section_based.py
```diff
@@ -234,37 +236,52 @@ def get_configuration_fields(self):
...
+        confirmed_counter = Counter()
+        requested_counter = Counter()
+        for p in participations:
+            if p.state == AbstractParticipation.States.CONFIRMED:
+                c = confirmed_counter
+            elif p.state == AbstractParticipation.States.REQUESTED:
+                c = requested_counter
+            else:
+                continue
+            c[p.data.get("dispatched_section_uuid", NO_SECTION_UUID)] += 1
```
jeriox, line 253: requested participations are put into the section "" rather than "other". This makes the whole counting process go wrong and the requested_count in the resulting signup stats is always 0

jeriox (changes requested):

ephios/settings.py
```diff
@@ -229,7 +229,7 @@
 SELECT2_I18N_PATH = ""
 
 # django-debug-toolbar
-if DEBUG:
+if DEBUG and env.bool("DEBUG_TOOLBAR", True):
```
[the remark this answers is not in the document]
felixrindt: Disabling the toolbar while developing without needing to change the settings file.
I wanted one because it was making my dev server very slow while working on the templates.

ephios/plugins/basesignup/signup/section_based.py
felixrindt: Very good catch, though I think the defect was a littler different/not only this. See a8f7fba for the refactor.

ephios/core/templates/core/home.html
```diff
@@ -52,7 +52,10 @@ <h5 class="card-title mb-0">{% trans "Events with requested participations" %}</
                     {% for shift in shifts %}
                         <li class="list-group-item flex-column">
                             <div class="d-flex w-100 justify-content-between align-items-center">
-                                <span>{{ shift.event.title }} ({{ shift.get_start_end_time_display }})</span>
+                                <span>
+                                    <span class="badge bg-warning">{{ shift.request_count }} {% translate "requests" %}</span>
```
felixrindt, line 56: What do you think about this?

ephios/plugins/basesignup/templates/basesignup/fragment_state_common.html
```diff
@@ -4,14 +4,18 @@
...
     {% endfor %}
 </div>
 
-<span data-bs-toggle="tooltip" data-bs-placement="top"
-      title="{% translate 'Needed' %}: {{ shift.signup_method.get_participation_number_info }}"
-><strong>{{ confirmed_count }}</strong> {% translate "in total" %}</span>
+{% with stats=shift.signup_method.get_signup_stats %}
+    <span class="pe-1">
+        <strong>{{ stats.confirmed_count }}{% if stats.max_count %}/{{ stats.max_count }}{% endif %}</strong>
+        {% translate "Total" context "participation count" %}
```
felixrindt, line 15: I started making use of context in the translation system, here in the template and in python using `pgettext`. Make sure you know what that is ([Docs](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#contextual-markers)) and respect it when translating.

jeriox: i know that, will be interesting to see if weblate knows it :P

ephios/plugins/basesignup/signup/section_based.py
jeriox: well but as the stats have been summed up before i think just fixing this line should have solved it?

ephios/core/templates/core/home.html
jeriox: I think it's too intrusive. If there are multiple events in the list there is a lot of yellow on the screen, and there is no reason to highlight every single entry because they all need similar attention regardless of the count of requests.
I think i'd prefer finding a small indicator (such as the number in the round label that you proposed before) that we can use here and on the event list to have a common look across the views

felixrindt: will remove

jeriox (approved):

PR #519, opened 2021-07-20
don't send notifications for own actions, add urls for push notifications

### LINKED ISSUE #416: Don't send notifications for own actions
As a user requesting participation for an event I'm responsible for,
I don't want to receive a notification about it,
so I don't waste my time reading about something I already know.

### LINKED ISSUE #518: Push notification target urls
As a user receiving PWA push notifications for events, 
I want to be able to click on the notification and be directly directed to the event details page so that it's easier to engage with the event.

jeriox:
closes #416 
closes #518


ephios/core/services/notifications/types.py
```diff
@@ -239,16 +239,20 @@ def send(cls, participation: AbstractParticipation):
...
-            notifications.append(
-                Notification(
-                    slug=cls.slug,
-                    user=user,
-                    data=dict(
-                        participation_id=participation.id,
-                        disposition_url=disposition_url,
-                    ),
+            if (
+                not participation.get_real_instance_class() == LocalParticipation
```
felixrindt, line 243: don't they recommend `is_instance` for this?

ephios/core/views/accounts.py
```diff
@@ -134,7 +134,8 @@ def post(self, request, *args, **kwargs):
                     name=self.object.get_full_name(), user=self.object
                 ),
             )
-            ProfileUpdateNotification.send(userprofile)
+            if request.user != userprofile:
+                ProfileUpdateNotification.send(userprofile)
```
felixrindt, line 138: might want to inform the user about password changes, even if it was done through their account

ephios/core/services/notifications/types.py
jeriox: i remember that we already had this discussion once. We have another spot in the code where we use this and I couldn't find smth in the docs

ephios/core/views/accounts.py
jeriox: password changes cant't happen via this view

felixrindt (approved):

PR #525, opened 2021-08-01
fix wrong select2 event, fix missing select2 translations

### LINKED ISSUE #143: Select2 widgets not translated
Select2 messages ("no results", ...) are always in english

### LINKED ISSUE #516: Select2 on disposition view shows error alert
Closing the select2 field in the disposition view without selecting an entry (e.g. by typing something into the field an then clicking somewhere outside the field) also triggers the form submission. This fails because no valid user has been selected and consequently shows an ugly alert to the user.

jeriox:
closes #143 
closes #516

### DIFF
--- a/ephios/templates/base.html
+++ b/ephios/templates/base.html
@@ -54,7 +54,7 @@
         <script type="text/javascript" src="{% static "bootstrap/js/bootstrap.bundle.min.js" %}"></script>
         <script type="text/javascript" src="{% static "select2/js/select2.js" %}"></script>
         <script type="text/javascript" src="{% static "django_select2/django_select2.js" %}"></script>
-        <script type="text/javascript" src="{{ select2_translation_url }}"></script>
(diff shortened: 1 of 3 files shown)


ephios/templates/base.html
```diff
@@ -54,6 +54,7 @@
         <script type="text/javascript" src="{% static "bootstrap/js/bootstrap.bundle.min.js" %}"></script>
         <script type="text/javascript" src="{% static "select2/js/select2.js" %}"></script>
         <script type="text/javascript" src="{% static "django_select2/django_select2.js" %}"></script>
+        <script type="text/javascript" src="{{ select2_translation_url }}"></script>
```
felixrindt, line 57:
```suggestion
        <script type="text/javascript" src="{% static ""select2/js/i18n/"|add:LANGUAGE_CODE|add:".js" %}"></script>
```

felixrindt: can you try this? If it works you don't need the `select2_translation_url` I think.

felixrindt (approved):

PR #560, opened 2021-09-23
Create SECURITY.md

felixrindt:
@jeriox can we create/do we have this mail address?


jeriox:
> 
> 
> @jeriox can we create/do we have this mail address?

yes we do

jeriox:
would also be great to have https://securitytxt.org/

jeriox (approved):

PR #572, opened 2021-10-03
fix rounding error in workinghour list

jeriox:
workinghour entries have been rounded to full hours before


ephios/core/models/users.py
```diff
@@ -158,18 +159,16 @@ def get_workhour_items(self):
...
-                / 3600,  # convert microseconds to seconds to hours
+                    output_field=models.DurationField(),
+                ),
                 date=ExpressionWrapper(TruncDate(F("shift__start_time")), output_field=DateField()),
                 reason=F("shift__event__title"),
             )
             .values("hours", "date", "reason")
         )
         workinghours = self.workinghours_set.all().values("hours", "date", "reason")
         hour_sum = (participation.aggregate(Sum("hours"))["hours__sum"] or 0) + (
```
felixrindt, line 170: Can you rename `participation` to `participations` and `hours` to duration or something like that?

PR #586, opened 2021-10-28
Fix inactive events showing in list

felixrindt:


ephios/core/models/events.py
```diff
@@ -73,7 +73,7 @@ class Meta:
         permissions = [("view_past_event", _("Can view past events"))]
         db_table = "event"
         base_manager_name = "all_objects"
-        default_manager_name = "all_objects"
+        default_manager_name = "objects"
```
jeriox, line 76: I think this could have some unwanted sideeffects, e.g. the django documentation states that this manager is used for the dumpdata management command. Are we sure we want it that way?

felixrindt: Well I think I would not want dumpdata to dump events that don't have shifts.

jeriox (approved):

PR #634, opened 2021-12-03
fix error page for HTTP500

jeriox:


felixrindt (approved):
It's *really* simple, but we can do something nicer later I guess.

PR #639, opened 2021-12-05
fix PWA orientation

### LINKED ISSUE #476: PWA does not respect orientation lock on Android

jeriox:
close #476


jeriox:
@felixrindt could you please validate that this works with your local setup? as the PWA is only served via HTTPS i cannot test this for myself

felixrindt (approved):
It did work, but I needed a PWA reinstall for it to take effect immediately.

PR #655, opened 2021-12-18
Make notification sending more robust

felixrindt:
This PR tries to:

* replace the `failed` field with something backend specific, so a fail in the email backend doesn't make the server not try webpush
* avoid problems when sent notifications are sent again should two instances of `run_periodic` running at the same time

Things missing:

* [ ] migration for removed `failed`


ephios/core/services/notifications/backends.py
```diff
@@ -25,25 +30,55 @@ def enabled_notification_backends():
...
 
+def send_a_notification(backend):
+    with transaction.atomic():
+        # grab an untried notification and mark it
+        if (
+            notification := Notification.objects.exclude(backends_tried__has_key=backend.slug)
+            .exclude(backends_tried__has_key=backend.slug)
+            .first()
+        ) is None:
+            return
```
felixrindt, line 41: Now this should only be None if another thread has snacked Notifications away, but if, we uselessly lock the database here.
Also, is there a nicer way in django to this "get me one and mark it as used" pattern?

Maybe we can use [`update`](https://docs.djangoproject.com/en/dev/ref/models/querysets/#update) with a limit=1?

jeriox: not sure if the update variant makes it more elegant as you'd need to make sure that you hit the right object within the limit=1

jeriox: do I miss smth or are the two exclude() calls exactly identical?

ephios/core/services/notifications/backends.py
```diff
@@ -25,25 +30,55 @@ def enabled_notification_backends():
...
+        notification.backends_tried[backend.slug] = False
+        notification.save()
+
+    # send it
+    if backend.can_send(notification) and backend.user_prefers_sending(notification):
+        try:
+            backend.send(notification)
+        except Exception as e:  # pylint: disable=broad-except
```
jeriox, line 49: we had some error mails for notifications where the corresponding object has been deleted in the meantime (e.g. notification for new event but the event has been deleted). Maybe we want to delete the notification immediately if a DoesNotExistExcpetion is thrown?

ephios/core/services/notifications/backends.py
```diff
@@ -25,25 +30,55 @@ def enabled_notification_backends():
...
+            if settings.DEBUG:
+                raise e
+            message = f"Notification {notification.pk}: {notification}\nException: {e}\n{traceback.format_exc()}"
```
jeriox, line 52: it would be great to have the SITE_URL in here as you cannot distinguish the source of the error by the sender anymore. currently this is only possible via the path in the stacktrace

ephios/core/services/notifications/backends.py
```diff
@@ -25,25 +30,55 @@ def enabled_notification_backends():
...
+            notification.errors[backend.slug] = message
+            notification.save()
+            mail_admins(f"Sending notification failed using {backend.slug}", message)
```
felixrindt, line 55: use logging instead?

ephios/settings.py
```diff
@@ -160,6 +160,10 @@
 LOGIN_REDIRECT_URL = "/"
 PASSWORD_RESET_TIMEOUT = 28 * 24 * 60 * 60  # seconds
 
+# Store notifications for 14 days after creation
+NOTIFICATION_RETENTION_SECONDS = 14 * 24 * 60 * 60
+NOTIFICATION_SENDING_PROCESS_COUNT = 1 if DEBUG else 4
```
felixrindt, line 165: should probably not have multiple processes try to send out emails using the same smtp server...

felixrindt:
We should probably, after not having touched this for a year, discuss what we can improve about notifications generally. Possible topics include:

* some "mark as read" functionality? We could have a notification history and if people open their email or click on a push notification or mark a notification as read manually show that.
* What is a `Notification` object? Currently, it ties a user to something that happened using multiple backends. Do we like this? Do we split it up?

jeriox:
superseeeded by #1143

Issue #557, opened 2021-09-19
Deal with uninstalled signup methods


felixrindt:
Lots of stuff breaks when a signup method is uninstalled. We could provide a fallback signup method that shows a warning (e.g. in render_shift_state) that the signup method is missing.

PR #667, opened 2022-01-05
Deal with missing/broken signup methods

felixrindt:
Fixes [#557](https://github.com/ephios-dev/ephios/issues/557)

* [x] catch errors of found signup methods
* [x] deal with missing signup methods
* [x] tests
* [x] get rid of this `report_exception` function and use proper logging instead - using the admin-mailer logging backend
* [ ] migrate #655 to use the new logging


ephios/core/templates/core/fragments/signup_method_missing.html
```diff
@@ -0,0 +1,11 @@
+{% load i18n %}
+
+<p class="text-muted">
+    {% translate "We're sorry! There is a problem displaying the state of the shift." %}
+</p>
+{% if request.user.is_superuser %}
+    <p class="text-muted">
+        Here's what we know:
+        {{ exception_message }}
```
jeriox, line 9: this would be a first. currently we don't display any exceptions inside the application, even for superusers. I think it is best practice to keep it that way

ephios/settings.py
```diff
@@ -196,26 +196,43 @@
...
+        "require_debug_false": {
+            "()": "django.utils.log.RequireDebugFalse",
         },
-        "loggers": {
-            "django": {
-                "handlers": ["mail_admins"],
-                "level": "ERROR",
-                "propagate": False,
-            },
+        "require_debug_true": {
```
jeriox, line 209: do we need this one?

ephios/core/templates/core/fragments/signup_method_missing.html
felixrindt: Oh yeah I wanted to remove that anyway. For the reasons you state.

ephios/core/models/events.py
```diff
@@ -282,7 +284,10 @@ def signup_method(self) -> "BaseSignupMethod":
         try:
             return signup_method_from_slug(self.signup_method_slug, self, event=event)
         except ValueError:
-            return None
+            logger.warning(f"signup method {self.signup_method_slug} was not found")
+            from ephios.core.signup.fallback import FallbackSignupMethod
+
+            return FallbackSignupMethod(self, event=event)
```
felixrindt, line 290: maybe, to be more semantically coherent, we should return `None` if the slug on the shift is None.

felixrindt: Meh it breaks the interface.

ephios/settings.py
felixrindt: not the `true` one, but I included both for completeness...

jeriox (approved):

PR #716, opened 2022-04-29
add poetry-dynamic-versioning-plugin

### LINKED ISSUE #638: Derive version from git tag
To prevent version bump commits it would be great to determine the target version from the current git tag. Poetry doesn't support that natively, but it will be possible with poetry 1.2 once it is released. Someone is building a plugin for that, the progess is tracked in https://github.com/mtkennerly/poetry-dynamic-versioning/issues/39

jeriox:
closes #638


.github/workflows/tests.yml
```diff
@@ -58,6 +60,8 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
     - uses: Gr1N/setup-poetry@v7
+      with:
+        poetry-preview: true
```
felixrindt, line 64: do we still need this?

felixrindt: we decided to wait for poetry 1.2 to be released

jeriox:
@felixrindt I removed the ephios dependency from ephios-testplugin for now, which solves all our problems

felixrindt (approved):

PR #726, opened 2022-05-25
Create API

felixrindt:
For now, lets build a simple (read-only) API for our most important resources (event, ...) using [Django Rest Framework](https://www.django-rest-framework.org/api-guide/serializers/).

* [x] Authentication: create token based auth for client applications? stuck to a user model? split account and user model? -> moved to #737
* [x] Permissions: integrate guardian
* [x] versioning?! -> later
* [x] tests
* [x] integrate django-filter
* [x] documentation? as part of the docs build process, generate the api schema and render docs for sphinx

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,3 +1,3 @@
 sphinx-github-changelog
 sphinxcontrib-openapi
-mistune<2.0.0  # pinned because of https://github.com/miyakogi/m2r/issues/66
+mistune<2.0.0  # June 2022: pinned because of https://github.com/miyakogi/m2r/issues/66
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,1 +1,2 @@
 sphinx-github-changelog
+sphinxcontrib-openapi
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
 sphinx-github-changelog
 sphinxcontrib-openapi
+mistune<2.0.0  # pinned because of https://github.com/miyakogi/m2r/issues/66
--- a/ephios/api/events.py
+++ b/ephios/api/events.py
@@ -86,6 +86,3 @@ class EventViewSet(viewsets.ReadOnlyModelViewSet):
         .prefetch_related(Prefetch("shifts__participations"))
         .order_by("start_time")
     )
-
-    def get_queryset(self):
-        return self.queryset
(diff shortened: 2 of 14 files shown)


felixrindt:
image

ephios/api/events.py
```diff
@@ -0,0 +1,89 @@
...
+    # pylint: disable=abstract-method
+    requested_count = serializers.IntegerField()
+    confirmed_count = serializers.IntegerField()
+    missing = serializers.IntegerField()
+    free = serializers.IntegerField()
+    min_count = serializers.IntegerField()
+    max_count = serializers.IntegerField()
+
+
+class ShiftSerializer(serializers.HyperlinkedModelSerializer):
```
jeriox, line 19: why do you use a HyperlinkedModelSerializer here?

ephios/settings.py
```diff
@@ -296,3 +300,13 @@ def GET_SITE_URL():
     if site_url.endswith("/"):
         site_url = site_url[:-1]
     return site_url
+
+
+REST_FRAMEWORK = {
+    "DEFAULT_PERMISSION_CLASSES": ["rest_framework.permissions.DjangoModelPermissions"],
```
jeriox, line 306: didn't test it, but according to the note in https://www.django-rest-framework.org/api-guide/permissions/#djangoobjectpermissions we should be using https://github.com/rpkilby/django-rest-framework-guardian

felixrindt: So I got the view permissions working. Here's the tricky details:

* DjangoModelPermissions checks for Model permissions. That can't be what we want, as we are also using object permissions.

* DjangoObjectPermissions checks if a user has model **and** object permissions, but not for **viewing**. Doesn't sound like what we want, but it actually is. See the next point.

* The CustomDjangoObjectPermissions shown in the django-rest-framework-guardian readme adds view permissions to the default ObjectPermission by DRF. Point is though, we don't normally give users view permissions for the event model. They wouldn't see the events they have object permissions for. To still achieve correct filtering, we use the DRF-guardian filter on the view.

docs/requirements.txt
```diff
@@ -1 +1,3 @@
 sphinx-github-changelog
+sphinxcontrib-openapi
+mistune<2.0.0  # pinned because of https://github.com/miyakogi/m2r/issues/66
```
felixrindt, line 3:
```suggestion
mistune<2.0.0  # June 2022: pinned because of https://github.com/miyakogi/m2r/issues/66
```

jeriox (approved):

PR #765, opened 2022-07-13
Dependency upgrades

felixrindt:
We had trouble upgrading our production environments:

- the environment existed for a while and had version 1 of a dependency installed.
- dependabot came along and upgraded the dependency to version 2. We installed that in the dev environment from `poetry.lock` and where dependent on features of that update, the requirements in `pyproject.toml` stayed untouched
- on the production environment, version 1 of the dependency was still installed and when upgrading the project with pip, pip's new default upgrade strategy `only-if-needed` didn't upgrade the dependency, because of course it wasn't needed according to `pyproject.toml`. 
- Now the failure occurred: version 1 was installed, but version 2 was required by the project.

This PR now upgrades all requirements in `pyproject.toml` and adds a few (hopefully all) missing ones (at least those that led to fails as described above). We will take a close look on our dependency declarations in the future.

Unfortunatly, we don't see a way to test for outdated minimum requirements automatically, as package managers don't provide a way to install the "minimum" required versions, right? And we can't keep old environments around in the CI pipeline...

Next time, when upgrading, consider to use pip's `--upgrade-strategy eager`.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -8,7 +8,7 @@ python-versions = "*"
 
 [[package]]
 name = "asgiref"
-version = "3.5.0"
+version = "3.5.2"
 description = "ASGI specs, helper code, and adapters"
 category = "main"
 optional = false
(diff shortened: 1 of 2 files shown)


jeriox:
command to update all dependencies to the latest version in pyproject.toml: https://github.com/python-poetry/poetry/issues/461#issuecomment-849610827

jeriox (approved):

PR #767, opened 2022-07-16
move notification settings to settings area, rework settings permission checking

### LINKED ISSUE #399: Permission granting for administration section
Currently only users with the `is_staff` flag set to true are able to access the "administration/settings" section. On the other hand, there are no UI controls to set this flag. There should either be a checkbox on the userprofile form or (preferrably) the ability to access the section should be allowed for members of the management group(s)

jeriox:
closes #399


jeriox:
@felixrindt I think I included everything from Arbeitspaket 1 apart from moving the password change view. For that, we are currently using the built-in django views with a custom template. To include that, we'd need to write our own view as it would need to inherit from our SettingsViewMixin. Should I do that?

ephios/core/views/settings.py
```diff
@@ -1,28 +1,36 @@
...
+    sections = []
+    if request.user.is_staff:
+        sections.append(
+            {
+                "label": _("ephios instance"),
+                "url": reverse("core:settings_instance"),
+                "active": request.resolver_match.url_name == "settings_instance",
+            }
+        )
+    if request.user.has_perm("core.view_eventtype"):
```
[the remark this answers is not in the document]
felixrindt: I mean it probably doesn't matter, but people will get 403s if they click on one of the buttons and they don't have permission to change.

ephios/core/views/settings.py
```diff
@@ -46,4 +54,30 @@ def form_valid(self, form):
...
+        return reverse("core:settings_notifications")
+
+    def form_valid(self, form):
+        form.update_preferences()
+        return super().form_valid(form)
+
+
+class PasswordChangeSettingsView(PasswordChangeView, SettingsViewMixin):
+    template_name = "core/settings/password_change_form.html"
+    success_url = reverse_lazy("core:settings_password_change_done")
```
felixrindt, line 79: image

The management menu entries are missing in this view.

ephios/core/signals.py
```diff
@@ -40,7 +40,8 @@
 administration_settings_section = PluginSignal()
 """
 This signal is sent out to get sections for administration settings. Receivers should return a list of dicts
-containing key-value-pairs for 'label', 'url' and a boolean flag 'active'.
+containing key-value-pairs for 'label', 'url' and a boolean flag 'active'. Only views that the current user is
+allowed to view should be returned.
```
felixrindt, line 44: Remember to update the beachguard plugin as well

ephios/core/views/settings.py
```diff
@@ -46,4 +54,30 @@ def form_valid(self, form):
...
+
+
+class PasswordChangeDoneSettingsView(PasswordChangeView, SettingsViewMixin):
+    template_name = "core/settings/password_change_done.html"
```
felixrindt, line 83: Can't we get rid of this view entirely?

ephios/core/urls.py
```diff
@@ -45,7 +44,12 @@
 )
 from ephios.core.views.log import LogView
 from ephios.core.views.pwa import manifest, offline, serviceworker
-from ephios.core.views.settings import GeneralSettingsView
+from ephios.core.views.settings import (
+    GeneralSettingsView,
```
felixrindt, line 48: Can you rename that class as well?

ephios/core/templates/core/settings/settings_base.html
```diff
@@ -12,12 +12,20 @@ <h1>{% translate "Settings" %} {% block settings_title %}{% endblock %}</h1>
...
+                           href="{% url "core:settings_password_change" %}">{% translate "Change password" %}</a></li>
+                {% if settings_sections %}
+                    <li class="nav-item mt-2"><h5>{% translate "Management" %}</h5></li>
                     {% for section in settings_sections %}
                     <li class="nav-item">
                         <a class="nav-link {% if section.active %}active{% endif %}"
                            href="{{ section.url }}">{{ section.label }}</a>
                     </li>
                     {% endfor %}
+                {% endif %}
```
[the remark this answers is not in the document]
felixrindt: Anyway, the signal name and the function name collecting the menu entries should use the same name, `management` or `administration`.

ephios/core/views/settings.py
jeriox: You should display some kind of confirmation to the user that the action worked, how would you do that otherwise?

ephios/core/views/settings.py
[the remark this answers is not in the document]
jeriox: Well the view that is behind the URL requires the view permission, so I thought it would be a good idea to check for that permission. Maybe there will be some setup where it could make sense to only allow people to view those types? For the 403 argument it would be better to hide the buttons in the list view if the user doesn't have the appropriate permissions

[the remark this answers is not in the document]
felixrindt: I think we can leave it for now and hide the buttons if the situation ever comes up.

ephios/core/views/settings.py
felixrindt: The messages framework.

ephios/core/templates/core/settings/settings_notifications.html
```diff
@@ -1,22 +1,10 @@
...
 {% load crispy_forms_filters %}
 {% load i18n %}
+{% load webpush_notifications %}
 
-{% block title %}
-    {% translate "Settings" %}
-{% endblock %}
-
-{% block html_head %}
-    {% webpush_header %}
```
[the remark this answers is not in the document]
jeriox: whoopsie

ephios/core/views/settings.py
jeriox: wrong inheritance order

ephios/core/views/settings.py
jeriox: good idea, thx

ephios/core/views/settings.py
```diff
@@ -31,14 +31,16 @@ def get_available_administration_settings_sections(request):
...
-    for __, result in administration_settings_section.send(None, request=request):
+    for __, result in management_settings_sections.send(None, request=request):
         sections += result
     return sections
 
 
 class SettingsViewMixin(FormView if typing.TYPE_CHECKING else object):
     def get_context_data(self, **kwargs):
-        kwargs["settings_sections"] = get_available_administration_settings_sections(self.request)
+        kwargs["management_settings_sections"] = get_available_administration_settings_sections(
```
felixrindt, line 41: would be great to rename `get_available_administration_settings_sections` as well

felixrindt (approved):

ephios/core/views/settings.py
jeriox: ah missed that

PR #772, opened 2022-07-17
Allow users with management permissions to make events visible for all groups

### LINKED ISSUE #771: Allow users with management permissions to make events visible for all groups
Allow users with management permissions to make events visible for all groups by attributing the global permission instead of an object-level permission. Make sure that all views respect the global permission.
Also reword the label of the management checkbox on the GroupForm, as we are now using "ephios instance" as a label for the settings that require the is_staff flag

jeriox:
closes #771

### DIFF
--- a/ephios/core/forms/users.py
+++ b/ephios/core/forms/users.py
@@ -52,6 +52,7 @@ MANAGEMENT_PERMISSIONS = [
     "core.add_qualification",
     "core.change_qualification",
     "core.delete_qualification",
+    "auth.publish_event_for_group",
     "modellogging.view_logentry",
 ]
 
@@ -118,7 +119,7 @@ class GroupForm(PermissionFormMixin, ModelForm):
         required=False,
     )
     is_management_group = PermissionField(
-        label=_("Can manage ephios"),
+        label=_("Can manage permissions and qualifications"),
         help_text=_(
             "If checked, users in this group can manage users, groups, all group memberships, eventtypes and qualifications"
         ),
--- a/ephios/core/forms/users.py
+++ b/ephios/core/forms/users.py
@@ -31,7 +31,7 @@ from ephios.extra.widgets import CustomDateInput
 from ephios.modellogging.log import add_log_recorder
 from ephios.modellogging.recorders import DerivedFieldsLogRecorder
 
-MANAGEMENT_PERMISSIONS = [
+CORE_MANAGEMENT_PERMISSIONS = [
     "auth.add_group",
     "auth.change_group",
     "auth.delete_group",
@@ -123,7 +123,7 @@ class GroupForm(PermissionFormMixin, ModelForm):
         help_text=_(
             "If checked, users in this group can manage users, groups, all group memberships, eventtypes and qualifications"
         ),
-        permissions=MANAGEMENT_PERMISSIONS,
+        permissions=CORE_MANAGEMENT_PERMISSIONS,
         required=False,
     )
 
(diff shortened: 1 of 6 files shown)


ephios/core/forms/users.py
```diff
@@ -52,6 +52,7 @@
     "core.add_qualification",
     "core.change_qualification",
     "core.delete_qualification",
+    "auth.publish_event_for_group",
```
felixrindt, line 55: I'm surprised to see the `auth.` here. It's because it's a permissions groups have on other groups, and the Group model is based in django's auth app?

ephios/core/forms/users.py
```diff
@@ -52,6 +52,7 @@
...
     "modellogging.view_logentry",
 ]
```
felixrindt, line 57: You could check of 

MANAGEMENT_PERMISSIONS zu CORE_MANAGEMENT_PERMISSIONS umbenennen (reines refactoring)

right here

tests/core/test_event_form.py
```diff
@@ -83,3 +84,13 @@ def test_participating_users_can_see_otherwise_invisible_event(
...
+
+
+def test_management_group_can_make_event_visible_for_all_groups(
+    django_app, service_event_type, manager, groups
+):
+    event_form = django_app.get(
+        reverse("core:event_create", kwargs=dict(type=service_event_type.pk)),
+        user=manager,
+    ).form
+    assert len(Group.objects.all()) == len(event_form["visible_for"].options)
```
felixrindt, line 96:
```suggestion
    assert set(Group.objects.all()) == set(event_form["visible_for"].options)
```

felixrindt (changes requested):
Please make the tests green and I'll have another look. Looks like some other tests need to be changed, I hope it didn't actually break something.

ephios/core/forms/users.py
jeriox: exactly

tests/core/test_event_form.py
jeriox: E       AssertionError: assert {<Group: Managers>, <Group: Planners>, <Group: Volunteers>} == {('2', True, 'Planners'), ('3', True, 'Volunteers'), ('1', True, 'Managers')}

jeriox:
> Please make the tests green and I'll have another look. Looks like some other tests need to be changed, I hope it didn't actually break something.

I removed the failing test as it was testing for behaviour that we don't want anymore

tests/core/test_event_form.py
felixrindt: ah whoopsie of course. I'd still like to check that semantically, the groups are the same

felixrindt, line 96: This assertion might be true because object permissions for all groups exist. Please make the test check that it's actually the general permission that makes this work.

felixrindt (approved):

ephios/core/forms/users.py
```diff
@@ -118,11 +119,11 @@ class GroupForm(PermissionFormMixin, ModelForm):
         required=False,
     )
     is_management_group = PermissionField(
```
felixrindt, line 121: I think, as we are disabling/force-checking the other checkboxes when this is checked, we should also hide the group selection for "allow publish for" when `is_management_group` gets checked.

felixrindt (changes requested):

ephios/core/forms/users.py
jeriox: I opted to hide it instead of disabling it to avoid confusion as there aren't all groups selected (which is not what actually happens permission-wise) but the field wouldn't accept changes

felixrindt (approved):

PR #781, opened 2022-08-09
GH Actions: upgrade ubuntu version, use inbuilt resources

felixrindt:
https://github.blog/changelog/2022-08-09-github-actions-ubuntu-22-04-is-now-generally-available-on-github-hosted-runners/

1. switch to 22.04 for testing
2. switch to ubuntu-latest for merge

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: [ 3.8 ]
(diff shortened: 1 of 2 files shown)


jeriox (approved):

jeriox:
as a sidenote: while reading the linked docs, i just stumbled across the fact that setup-python includes support for caching poetry dependencies, so we could get rid of our own caching solution as it already caused problems in the past.
https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

jeriox:
and another sidenote: the runners also include a postgres and a mysql service, so we could also use the pre-installed one instead of our own solution
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#databases

PR #845, opened 2022-12-28
fix entrypoint with PluginConfig

jeriox:


felixrindt (approved):

felixrindt:
Remember to update the cookie cutter template and existing non-base plugins

Issue #577, opened 2021-10-10
Change event type in event form


felixrindt:
As a planner I want to change the event's type after creation so I can correct a mistake.

Issue #472, opened 2021-05-07
Improve calendar design


jeriox:
As a user, I expect the event calendar view to display the shifts in small boxes with times inside of each calendar day (similiar to Google Calendar etc.)

jeriox:
Inspiration Moodle
![grafik](https://user-images.githubusercontent.com/12702135/178154982-f5bb0ed0-8bd7-415f-83bc-b09b94f1ed4b.png)
Inspiration Google Calendar
![grafik](https://user-images.githubusercontent.com/12702135/178155049-38fac8f9-3e45-43a0-b6fd-fb64b02c92ae.png)
Inspiration OWA
![grafik](https://user-images.githubusercontent.com/12702135/178155179-6a2485b5-082e-4d62-beec-c3f8f6a66e3c.png)


Besides using boxes and truncating the event title, I think we should also use a fixed height for the rows (=weeks). we could use something like the "+1" in OWA or "+1 more" in GCal if not all events fit inside the row. Another thing to think about is whether we want to display multiple shifts that occur on the same day (e.g day and night shift would currently result in two calender entries for the day)

Issue #418, opened 2021-04-05
Free object search for log view


felixrindt:
Followup from #404

As a viewer of the log view
I want to filter for specific objects from one of multiple types
so that I can filter down the results to only those that are of interest to me.

jeriox:
As this is quite complicated to achieve and there seems to be little benefit as there is a more intuitive path to see the desired results (navigating to the event and clicking on "history"), we won't implement this

PR #846, opened 2022-12-30
Event Filter

felixrindt:
* [x] rebuild calendar view
* [x] fix TODOs
* [x] check API compatibility
  * [x] while at it, maybe add a limit to the events API endpoint.
* [x] Pagination
* [x] remove past event view
* [x] calendar view must be scrolled on mobile if filter is always visible... what to do?
* [x] update and add tests

move to different issue/PR

* match this filters design with the modellogging filter -> seperate issue #860
  * specifically: the select2 inputs don't look like native bootstrap5 selects, so theres some theming broken
* remove past event permission -> seperate make-migrations PR
* #472 
* why can't I change the event type of an event later on? #577


jeriox:
regarding the modellogging filter you could also think about https://github.com/ephios-dev/ephios/issues/418

.env.example
```diff
@@ -11,3 +11,4 @@ EMAIL_URL=dummymail://
 DEFAULT_FROM_EMAIL=webmaster@localhost
 SERVER_EMAIL=root@localhost
 ADMINS=Root User <root@localhost.dev>
+# VAPID_PRIVATE_KEY_PATH=data/private_key.pem # push notifications, create in `data/` with `vapid --gen`
```
felixrindt, line 14: placeholder comment: remember to add the issues from the PR description

jeriox: the last one already exists (#577)

jeriox:
grafik
when first opening the calender view, I find it quite unintuitive where to look for the navigation. I expected them to be visible somewhere and needed to remind myself that they are inside of the filter. I think users will struggle with that, especially since they have been existing before

jeriox (approved):

PR #856, opened 2023-01-02
add selectable icon to simpleresource plugin

jeriox:


ephios/plugins/simpleresource/models.py
```diff
@@ -5,6 +5,7 @@
 
 class ResourceCategory(models.Model):
     name = models.CharField(max_length=50)
+    icon = models.CharField(max_length=50, default="box")
```
felixrindt, line 8: can you add a help text mentioning where to pick icon names from?

felixrindt (approved):

PR #858, opened 2023-01-09
add general required qualifications to eventtype

### LINKED ISSUE #588: General required Qualification
Build a plugin (or in core as a checker in the BaseSignupMethod) that lets you set required qualification per event type.
These will then always be checked.

Could be used for "Führungszeugnis"/"Hygieneeinweisung"/ etc...

jeriox:
closes #588

### DIFF
--- a/ephios/core/signup/methods.py
+++ b/ephios/core/signup/methods.py
@@ -379,6 +379,17 @@ def check_participant_age(method, participant):
         )
 
 
+def check_general_required_qualifications(method, participant):
+    if not participant.has_qualifications(
+        method.shift.event.type.preferences["general_required_qualifications"]
+    ):
+        return ParticipantUnfitError(
+            _("You lack qualifications for the eventtype {eventtype}.").format(
+                eventtype=method.shift.event.type
+            )
+        )
+
+
 def check_participant_signup_signal(method, participant):
     errors = []
     for _, result in check_participant_signup.send(None, method=method, participant=participant):
@@ -458,6 +469,10 @@ class BaseSignupMethodConfigurationForm(forms.Form):
         initial=True,
     )
 
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop("event", None)
+        super().__init__(*args, **kwargs)
+
 
 class BaseSignupMethod:
     # pylint: disable=too-many-public-methods
@@ -527,6 +542,7 @@ class BaseSignupMethod:
             check_inside_signup_timeframe,
             check_conflicting_participations,
             check_participant_age,
+            check_general_required_qualifications,
             check_participant_signup_signal,
         ]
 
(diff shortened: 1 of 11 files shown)


ephios/core/signup/methods.py
```diff
@@ -379,6 +379,17 @@ def check_participant_age(method, participant):
         )
 
 
+def check_general_required_qualifications(method, participant):
+    if not participant.has_qualifications(
+        method.shift.event.type.preferences["general_required_qualifications"]
+    ):
+        return ParticipantUnfitError(
+            _("You lack qualifications for the eventtype {eventtype}.").format(
```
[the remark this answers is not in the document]
felixrindt: "event type" also seems like a somewhat technical term at this point, although we are introducing the term in the event filter. 

I guess I need to propose something at this point:

`You lack the necessary qualification to participate in "{{ event_type }}" type events.`
`Ihnen fehlt die nötige Qualifikationen um an Veranstaltungen der Art "{{ event_type}}" teilzunehmen.`

ephios/plugins/basesignup/signup/common.py
```diff
@@ -81,7 +81,9 @@ def _participant_checkers(self):
     @staticmethod
     def check_qualification(method, participant):
         if not participant.has_qualifications(method.configuration.required_qualifications):
-            return ParticipantUnfitError(_("You are not qualified."))
+            return ParticipantUnfitError(
+                _("You don't have all required qualifications for this shift.")
```
felixrindt, line 85: Again, I don't think the plural is nice. I like the differentiation for event type vs shift in the phrasing. Something like
`You lack the necessary qualification to participate in this shift.` might be better suited.

ephios/plugins/basesignup/signup/common.py
```diff
@@ -92,6 +94,18 @@ class ConfigurationForm(super().configuration_form_class):
                 widget=Select2MultipleWidget,
                 required=False,
                 initial=[],
+                help_text=_(
+                    "Participants also need to have the qualifications <b>{qualifications}</b> to participate in {eventtype}"
```
felixrindt, line 98: I like the `<b>` here.

ephios/core/signup/methods.py
```diff
@@ -458,6 +469,10 @@ class BaseSignupMethodConfigurationForm(forms.Form):
         initial=True,
     )
 
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop("event", None)
```
felixrindt, line 473: \*sigh*
I don't like that we need to introduce the event parameter here but I guess it makes sense. In the usage down below you just expect `self.event` to not be None, so this should really be:
```suggestion
    def __init__(self, *args, **kwargs):
        self.event = kwargs.pop("event")
```

felixrindt (approved):

PR #864, opened 2023-01-11
Rework event_list_signup_state_counts filter

### LINKED ISSUE #861: Consider removing event_signup_state_counts filter
With #846 we introduced counting participation states using the database, which might have made the filter `event_signup_state_counts` used in `event_list.html` obsolete. We should try to remove the filter and use the database-annotations instead. Beware though that currently, the annotation is only computed if the filter form is valid and needs them.

felixrindt:
Fixes #861

### DIFF
--- a/ephios/core/views/event.py
+++ b/ephios/core/views/event.py
@@ -139,7 +139,6 @@ class EventFilterForm(forms.Form):
             )
 
         if state_filter := fdata.get("state"):
-            qs = self._annotate_event_participation_states(qs)
             qs = {
                 "confirmed": qs.filter(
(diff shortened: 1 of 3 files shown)


ephios/core/views/event.py
```diff
@@ -105,16 +105,6 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["date"].initial = timezone.now().date()
 
-    def clean_date(self):
```
jeriox, line 108: is this removed on purpose?

jeriox (approved):

ephios/core/views/event.py
felixrindt: yeah we don't need it

PR #870, opened 2023-01-21
add signal for additional PermissionFields on GroupForm

### LINKED ISSUE #768: Allow plugins to customize GroupForm
Plugins können Gruppen-Settings erweitern um Permissions group-based festzulegen

- Pages-Plugin sollte das nutzen

jeriox:
closes #768

### DIFF
--- a/ephios/core/signals.py
+++ b/ephios/core/signals.py
@@ -128,6 +128,13 @@ This signal is sent out to get additional information to display on the homepage
 Receivers receive a ``request`` keyword argument. Receivers should return html that will be rendered inside a card.
 """
 
+register_group_permission_fields = PluginSignal()
+"""
+This signal is sent out to get a list of permission fields that should be displayed on the group form.
+Receivers should return a list of tuples of the form ``(field_name, field)``. field must be an instance of
+``ephios.extra.permissions.PermissionField``.
+"""
+
 
 @receiver(
     register_consequence_handlers,
--- a/ephios/extra/permissions.py
+++ b/ephios/extra/permissions.py
@@ -39,7 +39,7 @@ class PermissionField(BooleanField):
 
     def __init__(self, *args, **kwargs):
         self.permission_set = kwargs.pop("permissions")
-        super().__init__(*args, **kwargs)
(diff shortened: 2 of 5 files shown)


ephios/core/signals.py
```diff
@@ -128,6 +128,13 @@
 Receivers receive a ``request`` keyword argument. Receivers should return html that will be rendered inside a card.
 """
 
+register_group_permission_fields = PluginSignal()
+"""
+This signal is sent out to get a list of permission fields that should be displayed on the group form.
```
felixrindt, line 133: is it really permission fields only we expect here? Could be any field (though it wouldn't do anything I guess?!)

ephios/extra/permissions.py
```diff
@@ -39,7 +39,7 @@ class PermissionField(BooleanField):
 
     def __init__(self, *args, **kwargs):
         self.permission_set = kwargs.pop("permissions")
-        super().__init__(*args, **kwargs)
+        super().__init__(*args, required=kwargs.pop("required", False), **kwargs)
```
felixrindt, line 42: This makes it default to not being required... Which makes sense for this use case, but deviates from the default Boolean Field (which has other semantics...)

felixrindt (approved):

ephios/core/signals.py
jeriox: I remembered that we talked about adding PermissionFields via this signal because this is a common use case for plugins and keeps the interface simple. We argued that we can introduce something like the PluginForm on events and shifts if there is a need for pluigns to add custom forms. so i thought we agreed on only accepting permissionfields here

ephios/extra/permissions.py
jeriox: yes, that was my intention as I think it should be an explicit choice to mark this as required and plugins could easily forget the required=False especially if they don't have the background of the PermissionField inheriting from BooleanField

ephios/core/signals.py
felixrindt: Yeah thanks for explaining!

PR #881, opened 2023-02-02
Change event_action_signal interface

felixrindt:
I think this change makes the interface more flexible, as receivers can reverse the url themselves or do something else entirely or just not return an action if not appropriate for that event.


jeriox (approved):

jeriox:
> I think this change makes the interface more flexible, as receivers can reverse the url themselves or do something else entirely or just not return an action if not appropriate for that event.

yeah that was probably just copy and paste from the event_bulk_action where it makes sense that it is specified how the pks are transferred to the action. No one is currently using the event_action, right?

PR #883, opened 2023-02-03
allow for events to be cancelled

### LINKED ISSUE #373: Cancel event
As a planner, I expect the event detail page to have a button to cancel the event. I expect to be able to enter some text to explain the reason. I expect the event to be set to inactive, remove all participations and send a notification to all participants.

jeriox:
closes #373

### DIFF
--- a/ephios/core/forms/events.py
+++ b/ephios/core/forms/events.py
@@ -300,3 +300,26 @@ class EventNotificationForm(forms.Form):
         ):
             raise ValidationError(_("You cannot send an empty mail."))
         return super().clean()
+
+
+class EventCancellationForm(forms.Form):
+    explanation = forms.CharField(
+        required=False,
+        widget=forms.Textarea,
+        label=_("Explanation"),
+        help_text=_(
+            "All participants will be notified about the cancellation and this message will be included."
+        ),
+    )
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop("event")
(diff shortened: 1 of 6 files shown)


jeriox:
@felixrindt I'm not sure how we want to proceed with a delete button for the event. Should we always offer both options? Or only allow the event to be deleted without being cancelled first when there are no participations?

felixrindt:
> @felixrindt I'm not sure how we want to proceed with a delete button for the event. Should we always offer both options? Or only allow the event to be deleted without being cancelled first when there are no participations?

I think we should allow both, that is:

* deleting an event entirely, even if it has participations without sending an email
* the more graceful approach introduced in this PR.

Maybe we can join the two together into one form. Also, what about the bulk delete action? Should it work on events with participations?

jeriox:
> > @felixrindt I'm not sure how we want to proceed with a delete button for the event. Should we always offer both options? Or only allow the event to be deleted without being cancelled first when there are no participations?
> 
> I think we should allow both, that is:
> 
>     * deleting an event entirely, even if it has participations without sending an email
> 
>     * the more graceful approach introduced in this PR.
> 
> 
> Maybe we can join the two together into one form. Also, what about the bulk delete action? Should it work on events with participations?

I like that idea. The user could then only have one button with "delete event" and can check a checkbox in the delete view with "notify users about the cancellation". if checked, the textarea for the message appears.

I don't see a reason why the bulk action shouldn't work?

PR #902, opened 2023-03-03
update django-environ

jeriox:


felixrindt (approved):
sure. closes #900

jeriox:
yeah I thought it would be more obvious that this is an important change if it's not coming from dependabot and I did the commit on the other branch anyways but was too slow to get the push out before the auto-merge kicked in

PR #904, opened 2023-03-06
add oauth2 flow as provider

jeriox:
- [x] make frontend schick
    - [x] Authorize screen
    - [x] CRUD views für applications
    - [x] user token view
- [x] scopes definieren
- [x] bestehende endpoints updaten (scopes und permissions ranschreiben)
- [x] Was passiert beim Logout? Was passiert bei Passwort-Änderung?
- [x] restrict application creation to staff users
- [x] create followup issues:
    - [x] API-Key mit Gruppen permissions -> #906
    - [x] Idee: Manual User Token creation -> #907


ephios/core/services/password_reset.py
```diff
@@ -12,6 +12,11 @@
 class CustomMinimumLengthValidator(MinimumLengthValidator):
     def password_changed(self, password, user):
         if user is not None:
+            # revoke API tokens
+            for token in user.oauth2_provider_accesstoken.all():
```
jeriox, line 16: probably needs change to our own model

ephios/extra/permissions.py
```diff
@@ -1,10 +1,27 @@
...
+    def _wrapped_view(request, *args, **kwargs):
+        if request.user.is_staff:
+            return view_func(request, *args, **kwargs)
+        if not request.user.is_authenticated:
+            from django.contrib.auth.views import redirect_to_login
+
+            return redirect_to_login(request.get_full_path())
+        raise PermissionDenied
+
+    return _wrapped_view
```
felixrindt, line 22: I gather that `staff_member_required` redirects to the admin login and isn't the right choice here then?
Did you copy this code from somewhere?

ephios/core/views/settings.py
```diff
@@ -21,6 +21,13 @@ def get_available_management_settings_sections(request):
                 "active": request.resolver_match.url_name == "settings_instance",
             }
         )
+        sections.append(
+            {
+                "label": _("OAuth2 applications"),
+                "url": reverse("oauth2_provider:list"),
+                "active": request.resolver_match.url_name == "list",
```
felixrindt, line 28: I have the feeling we must be more specific with the `active` predicate here.

ephios/api/signals.py
```diff
@@ -0,0 +1,10 @@
+from django.dispatch import receiver
+
+from ephios.core.signals import periodic_signal
+
+
+@receiver(periodic_signal, dispatch_uid="ephios.api.signals.clear_expired_tokens")
+def clear_expired_tokens(sender, **kwargs):
+    from oauth2_provider.models import clear_expired
+
+    clear_expired()
```
felixrindt, line 10: For plugins, we import `signals.py` in the app config, but that's not happening with the `api` app. Please make sure this file is imported.

felixrindt: Why are we doing this? Wouldn't it be helpful to retain expired tokens for a few days for debugging/forensics?

ephios/settings.py
```diff
@@ -68,7 +69,7 @@
     "ephios.extra",
     "ephios.api",
 ]
-INSTALLED_APPS += EPHIOS_CORE_MODULES
+INSTALLED_APPS = EPHIOS_CORE_MODULES + INSTALLED_APPS
```
felixrindt, line 72: Why change the ordering?

ephios/settings.py
```diff
@@ -311,5 +312,25 @@ def GET_SITE_URL():
     "DEFAULT_AUTHENTICATION_CLASSES": [
         "rest_framework.authentication.TokenAuthentication",
         "rest_framework.authentication.SessionAuthentication",
+        "oauth2_provider.contrib.rest_framework.OAuth2Authentication",
     ],
 }
+
+# Like UserProfile, these models are implemented using djangos private swappable API
+# due to how swappable works, we need to list every model for it to work, even if we are not swapping all
```
felixrindt, line 320: update the comment please. We not only need to list every model, but swap every model to our own version.

felixrindt (changes requested):

ephios/extra/permissions.py
jeriox: exactly.
nope, self-written but I looked at user_passes_test to figure out how a decorator works ^^

ephios/core/views/settings.py
jeriox: do you have an idea how? I didn't wanna change the url_names of oauth2_provider

ephios/settings.py
jeriox: oauth2 allows to change the template, but the way they want us to do that is by overriding the oauth2_provider/*.html templates. the first match is used by django, so with the old ordering django always served the built-in template

ephios/core/views/settings.py
felixrindt: Resolver_match should also provided the app name via an attribute?!

ephios/api/signals.py
felixrindt: not yet. It's fine for now I guess. Still needs the signals import though.

ephios/api/templates/oauth2_provider/authorized-tokens.html
```diff
@@ -0,0 +1,23 @@
...
+                {{ authorized_token.application }}
+                (<a href="{% url 'oauth2_provider:authorized-token-delete' authorized_token.pk %}">{% trans "revoke" %}</a>)
+            </li>
+            <ul>
+            {% for scope_name, scope_description in authorized_token.scopes.items %}
+                <li>{{ scope_name }}: {{ scope_description }}</li>
+            {% endfor %}
+            </ul>
+        {% empty %}
+            <li>{% trans "There are no authorized tokens yet." %}</li>
```
felixrindt, line 19: I really don't like the look of this list or the other templates.

ephios/api/templates/oauth2_provider/authorize.html
```diff
@@ -0,0 +1,38 @@
...
+{% block content %}
+    {% if not error %}
+            <form id="authorizationForm" method="post">
+                <h3 class="block-center-heading">{% trans "Authorize" %} {{ application.name }}?</h3>
+                {% csrf_token %}
+
+                {% for field in form %}
+                    {% if field.is_hidden %}
+                        {{ field }}
+                    {% endif %}
```
felixrindt, line 13: Havn't looked at the authorize page yet but why are we rendering hidden fields here?

ephios/api/templates/oauth2_provider/authorized-tokens.html
felixrindt: Also, there should be some explanatory sentence as to what the page shows.

ephios/settings.py
```diff
@@ -311,5 +312,25 @@ def GET_SITE_URL():
     "DEFAULT_AUTHENTICATION_CLASSES": [
         "rest_framework.authentication.TokenAuthentication",
```
felixrindt, line 313: I think we should remove the Token Auth with this PR, as we really want to transition to our own Token Model. 

* remove Token Auth
* remove Token app
* migration (?)

felixrindt (changes requested):

ephios/api/templates/oauth2_provider/authorized-tokens.html
jeriox: I didn't want to reproduce them in the first place, but had to because of their inheritance structure. these are 1:1 copied from the library

ephios/api/templates/oauth2_provider/authorize.html
jeriox: they explain that in the documentation, they rely on that to carry the information through from the initial request to the validation of the form the user submits

felixrindt: can you add a comment pls?

ephios/api/templates/oauth2_provider/authorized-tokens.html
felixrindt: Want to split up the work on making the views look a little better?

ephios/settings.py
felixrindt: I see, thanks. Maybe put a comment so we wont break it when touching this in the future?

felixrindt (commented):
I guess for this PR only thing left (besides addressing the small review comments) is making the pages look a little better...

ephios/api/templates/oauth2_provider/authorized-tokens.html
jeriox: maybe @gritfessel wants to work on that after merging this PR?

ephios/settings.py
jeriox: something in the back of my head said that another dependency required us to import it _after_ the dependency, but I couldn't remember which one (my head said dynmic-preferences, but that wasn't true). As tests still work I would suggest going with this PR and coming back to it if something breaks

felixrindt (approved):

ephios/api/templates/oauth2_provider/authorized-tokens.html
jeriox: moved to #938

PR #909, opened 2023-03-10
Refactor: nullable date-of-birth

felixrindt:
In preparation for 3rd party auth and also to reduce the number of fields on saas registration, I propose to make the date of birth field on local users nullable. It can't be blank on the user edit form though, so we still enforce it for users created through the CRUD views.


ephios/core/models/users.py
```diff
@@ -123,13 +124,15 @@ def get_short_name(self):
         return self.first_name
 
     @property
-    def age(self):
+    def age(self) -> Optional[int]:
+        if self.date_of_birth is None:
+            return None
```
jeriox, line 129: I think this could go sideways in a few places where calculations are done on the age, as we expected it to be always  there. Did you check to see where this is used? (also thinking about the youthwarden plugin)

felixrindt: Yeah also looks fine...

jeriox (approved):

PR #950, opened 2023-05-19
add OIDC client

jeriox:


ephios/settings.py
```diff
@@ -335,3 +335,14 @@ def GET_SITE_URL():
...
+if ENABLE_OIDC_CLIENT := env.bool("ENABLE_OIDC_CLIENT", False):
+    INSTALLED_APPS.append("mozilla_django_oidc")
+    AUTHENTICATION_BACKENDS.append("mozilla_django_oidc.auth.OIDCAuthenticationBackend")
+    OIDC_RP_CLIENT_ID = env.str("OIDC_RP_CLIENT_ID")
+    OIDC_RP_CLIENT_SECRET = env.str("OIDC_RP_CLIENT_SECRET")
+    OIDC_RP_SIGN_ALGO = env.str("OIDC_RP_SIGN_ALGO")
+    OIDC_OP_AUTHORIZATION_ENDPOINT = env.str("OIDC_OP_AUTHORIZATION_ENDPOINT")
+    OIDC_OP_TOKEN_ENDPOINT = env.str("OIDC_OP_TOKEN_ENDPOINT")
+    OIDC_OP_USER_ENDPOINT = env.str("OIDC_OP_USER_ENDPOINT")
+    OIDC_OP_JWKS_ENDPOINT = env.str("OIDC_OP_JWKS_ENDPOINT")
```
felixrindt, line 348: that looks complicated to configure. I think we should create a docs page linking to mozillas OIDC documentation...

felixrindt (approved):

ephios/settings.py
jeriox: the problem is that all of these values need to come from the OP, so the docs there also say little apart from "use the values that you get from your OP"

felixrindt: Yeah I know. I just think we should create a docs page for these kind of features.

jeriox: done

PR #952, opened 2023-05-22
API/OAuth: Improve templates, add user tokens

felixrindt:


ephios/api/oauth2/views.py
```diff
@@ -0,0 +1,9 @@
+from oauth2_provider.models import get_application_model
+from oauth2_provider.views import ApplicationList
+
+from ephios.extra.mixins import StaffRequiredMixin
+
+
+class AllUserApplicationList(StaffRequiredMixin, ApplicationList):
```
jeriox, line 7: is this actually used as the URLs are not changed?

ephios/api/templates/oauth2_provider/application_list.html
```diff
@@ -3,18 +3,41 @@
...
+                    <tr>
+                        <td>{{ application.name }}</td>
+                        <td>
+                            {{ application.user }}
+                        </td>
+                        <td>{{ application.created }}</td>
+                        <td class="d-flex">
+                            <a class="btn btn-secondary"
+                               href="{{ application.get_absolute_url }}">{% trans "View" %}</a>
+                            <a class="btn btn-secondary ms-1"
```
jeriox, line 29: usually we have additional icons on the buttons in tables

ephios/api/templates/oauth2_provider/application_list.html
```diff
@@ -3,18 +3,41 @@
 {% load i18n %}
 {% block settings_content %}
     <div class="block-center">
-        <h3 class="block-center-heading">{% trans "Your applications" %}</h3>
+        <h3 class="block-center-heading">{% trans "OAuth2 applications" %}</h3>
+        <a class="btn btn-primary" href="{% url "oauth2_provider:register" %}">{% trans "New Application" %}</a>
         {% if applications %}
```
jeriox, line 8: I was asked what OAuth is and what you can configure here, that isn't obvious to most planners. A short paragraph here about what this page is for would be great

jeriox (approved):

ephios/api/templates/oauth2_provider/application_list.html
felixrindt: its meant for staff users, not planners, but we sure can add a description and call it "3rd party integration" or smth like that?!

PR #953, opened 2023-05-22
API: User token

### LINKED ISSUE #907: Allow manual creation of API access tokens for users
Currently, user can only obtain an access token for the API via an OAuth2 flow, which requires an application set up by an instance administrator. We could think about whether we want to allow users to create their own access tokens to be able to access their data via the API apart from OAuth apps

felixrindt:
* [x] add description field to AccessToken
* [x] Token creation workflow
* [x] admin view to revoke other users' tokens
* [x] sort/group tokens in token list
* [x] add tests for User Token views and permission checks
* [x] add documentation

closes #907


ephios/api/urls.py
```diff
@@ -15,4 +24,58 @@
...
+                    name="settings-oauth-app-register",
+                ),
+                path(
+                    "applications/<int:pk>/",
+                    staff_required(views.ApplicationDetail.as_view()),
+                    name="settings-oauth-app-detail",
+                ),
+                path(
+                    "applications/<int:pk>/delete/",
+                    staff_required(views.ApplicationDelete.as_view()),
```
jeriox, line 70: when deleting an application, the success url is wrong and results in a NoReverseMatch at /api/settings/applications/1/delete/

Reverse for 'list' not found. 'list' is not a valid view function or pattern name.

ephios/api/templates/api/access_token_list.html
```diff
@@ -0,0 +1,98 @@
...
+        {% endif %}
+
+        {% for authorized_token in authorized_tokens %}
+            <div class="card w-100 mb-3">
+                <div class="card-body">
+                    <div class="row">
+                        <div class="col-12 col-md-6">
+                            <div class="d-flex justify-content-between align-items-start">
+                                <h5 class="card-title">
+                                    {{ authorized_token }}
```
jeriox, line 31: this reads `Access Token #2`, not sure if this is so helpful

ephios/api/models.py
```diff
@@ -18,19 +21,55 @@ class Meta(AbstractApplication.Meta):
...
-    def name(self):
-        if self.application and self.application.name:
-            return f"{self.application.name} #{self.id}"
-        return _("User-Token") + f" #{self.id}"
+    description = models.TextField(
+        verbose_name=_("Description"),
+        blank=True,
+    )
+
+    revoked = models.DateTimeField(null=True)
```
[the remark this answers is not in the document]
jeriox: also, I think that revoked tokens won't be deleted by the cronjob

ephios/api/templates/api/access_token_list.html
```diff
@@ -0,0 +1,98 @@
...
+                If you are unsure about the origins of a token or the security of the third party application
+                please revoke the token.
+            {% endblocktranslate %}
+        </p>
+
+        <a href="{% url "api:settings-access-token-create" %}"
+           class="btn btn-secondary mb-3">{% translate "Create new API token" %}</a>
+
+        {% if not request.GET.show_inactive %}
+            <a href="?show_inactive=1" class="btn btn-secondary ms-1 mb-3">{% translate "Show inactive" %}</a>
```
jeriox, line 21: do you think that this will be helpful for users? I think that it may be too much

ephios/api/templates/oauth2_provider/application_confirm_delete.html
```diff
@@ -8,13 +8,13 @@ <h3 class="block-center-heading">
                 Are you sure to delete the application {{ app_name }}?
```
jeriox, line 8: app_name may be empty, it looks a bit weird then

jeriox (changes requested):

ephios/api/models.py
```diff
@@ -18,19 +21,55 @@ class Meta(AbstractApplication.Meta):
...
+            return _("Access Token") + f" #{self.id}"
+        return _("User Token") + f" #{self.id}"
+
+    def revoke_related(self):
+        """
+        Revoke this token and other types of related OAuth2 tokens.
+        """
+        try:
+            # aquired from OAuth2 refresh token, so revoke that one (which will revoke this access token as well)
+            self.refresh_token.revoke()
```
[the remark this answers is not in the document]
felixrindt: the `revoke` function on refresh tokens also revokes its corresponding access tokens.

ephios/api/migrations/0002_accesstoken_description_accesstoken_revoked_and_more.py
```diff
@@ -0,0 +1,27 @@
...
+        ),
+        migrations.AddField(
+            model_name="accesstoken",
+            name="revoked",
+            field=models.DateTimeField(null=True),
+        ),
+        migrations.AlterField(
+            model_name="accesstoken",
+            name="expires",
+            field=models.DateTimeField(null=True),
```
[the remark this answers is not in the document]
felixrindt: its non-nullable in OAuth2, we allow it to be null for our additional use case. I don't expect it to cause a problem...

ephios/api/models.py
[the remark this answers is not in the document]
felixrindt: I think semantically expiration and revocation are two different things and I wanted to have them seperate. 
I'll look into deleting tokens that are expired. Could then also remove the "view inactive" button...

ephios/api/models.py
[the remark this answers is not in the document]
jeriox: ah perfect

ephios/api/urls.py
felixrindt: damnit I'll just write test for these too

felixrindt: fixed

ephios/api/templates/api/access_token_list.html
```diff
@@ -4,95 +4,125 @@
...
+                            {{ token.description|rich_text }}
+                        </h5>
+                        <p class="text-muted">
+                            {% blocktranslate trimmed with created=token.created %}
+                                created on {{ created }}
+                            {% endblocktranslate %}
+                            <br/>
+                            {% if token.revoked %}
+                                {% blocktranslate trimmed with revoked=token.expires %}
+                                    revoked on {{ revoked }}
```
jeriox, line 83: this isn't a date anymore, is it?

jeriox (approved):

ephios/api/templates/api/access_token_list.html
felixrindt: its not, but the variable revoked in the translation is set to the expiration date used as surrugate

PR #957, opened 2023-05-25
staging build: fetch all commits

felixrindt:
The staging build is using dynamic versioning, but not setting the version correctly:

image

I'm guessing that's because GH actions only fetches the latest commit by default. So lets fix that!
(see https://github.com/actions/checkout#checkout-v3 )


jeriox (approved):
wow, really good catch!

Issue #357, opened 2021-03-07
Spinner and blurring for PWA does not work on iOS


jeriox:

felixrindt:
Also, when downloading a PDF on mobile, the page doesn't change. So when hitting the back button on my android phone, I get back to the still blurry ephios page.

felixrindt:
> Also, when downloading a PDF on mobile, the page doesn't change. So when hitting the back button on my android phone, I get back to the still blurry ephios page.

This also happens just hitting 'back' from a webpage. As the prev. page is not reloaded, the spinner does not disappear.

jeriox:
> > Also, when downloading a PDF on mobile, the page doesn't change. So when hitting the back button on my android phone, I get back to the still blurry ephios page.
> 
> This also happens just hitting 'back' from a webpage. As the prev. page is not reloaded, the spinner does not disappear.

this is fixed by #965, blurring still not working on iOS

PR #965, opened 2023-05-31
PWA: unblur spinner on pageshow

felixrindt:
happened when hitting the back button and it was bugging me so hard I fixed it tonight :muscle: 
These were the interesting reads:

* pointing me to the event that worked on pageload: https://christopheraue.net/design/fading-pages-on-load-and-unload
* more info on the page lifecycle and the chrome exclusive APIs if this fix does not work everywhere: https://developer.chrome.com/blog/page-lifecycle-api/

This is related to my comments on #357, but not the initial part about it not working on iOS. Maybe @jeriox can have another look?

### DIFF
--- a/ephios/static/ephios/js/main.js
+++ b/ephios/static/ephios/js/main.js
@@ -54,10 +54,16 @@ $(document).ready(function () {
     if (navigator.standalone || window.matchMedia('(display-mode: standalone)').matches) {
         $(window).on('beforeunload', function () {
             $('.blur-on-unload').addClass("unloading");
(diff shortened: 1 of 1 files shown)


jeriox:
omg great! will have a look on iOS once i have my iPad at hand

jeriox (approved):
problem with back navigation fixed, blur still not present on iOS

felixrindt:
Seems like it's a known "bug" in webkit (see https://bugs.webkit.org/show_bug.cgi?id=199854 )
I could not find a solution for iOS

PR #972, opened 2023-06-07
add federation

jeriox:
first draft with on-demand events fetching according to design blog.

- [x] frontend view that collects shared events from hosts
  - [x] make it look like the regular event view
  - [x] display correct signup url
- [x] views for OAuth flow
  - [x] transmit qualifications
  - [x] note that other inclusions may lead to weird stuff
- [x] event detail view for federated users
  - [x] decide how to authorize against host
  - [x] permission checking (is the event shared for the guest?)
- [x] shift signup for federated users
- [x] view to manage share for an event
  - [x] use checkboxmultiselect
- [x] invite-flow
  - [x] add explanations
  - [x] add delete views
  - [x] prevent OAuth app from being deleted
  - [x] handle invite code expiration
  - [x] confirmation view when redeeming code
  - [x] show invite code in reveal view
  - [x] invite code with URL as complete link
- [x] fix naming
  - Menüpunkt: Externe Veranstaltungen
  - Plugin: Federation
  - Eventform: Share event with other ephios instances
  - Settings: Ephios instances that you share events with; Ephios instances that share events with you: Share events with someone else; Enter invite key
- [x] check if own permission class for sharedeventlistview is a good idea (maybe set guest there?) -> doesn't make sense as the signature only allows to return a user and one cannot set stuff in the request
- [x] tests


ephios/plugins/federation/models.py
```diff
@@ -0,0 +1,90 @@
...
+
+class FederatedGuest(models.Model):
+    name = models.CharField(max_length=255)
+    url = models.URLField()
+    access_token = models.ForeignKey(AccessToken, on_delete=models.CASCADE)
+    client_id = models.CharField(max_length=255)
+    client_secret = models.CharField(max_length=255)
+
+
+class FederatedHost(models.Model):
```
felixrindt, line 19: Reading the model names causes a lot of confusion in me. Maybe we should name all models with clear terminologic prefix like `IncomingFederatedUser` and `OutgoingFederatedEventShare` and `OutgoingFederationPartnership`.

ephios/plugins/federation/views/api.py
```diff
@@ -28,3 +37,88 @@ def get_queryset(self):
...
+                    user.qualifications.add(Qualification.objects.get(uuid=qualification["uuid"]))
+                except Qualification.DoesNotExist:
+                    for included_qualification in qualification["includes"]:
+                        try:
+                            user.qualifications.add(
+                                Qualification.objects.get(uuid=included_qualification["uuid"])
+                            )
+                        except Qualification.DoesNotExist:
+                            continue
+        self.request.session["federated_user"] = user.pk
```
felixrindt, line 124: A qualification `includes` is only the direct includes, not the transitive ones, right? If so, I don't like that we only follow it for one layer here. It should be all including or None if the one couldn't be found. Only doing one round of breath first search might lead to behavior that confuses users.

jeriox: no, the serialization uses the results from collect_all_qualifications

ephios/api/templates/oauth2_provider/application_list.html
```diff
@@ -33,6 +33,13 @@ <h4>{% translate "OAuth2 applications" %}</h4>
                         <td>{{ application.created }}</td>
                         <td class="d-flex">
                             <div>
+                                {% if application.federatedhost_set.exists %}
+                                <a class="btn btn-secondary"
+                                   href="{% url "federation:settings" %}">
+                                    <span class="fa fa-cog"></span>
+                                    <span class="d-none d-lg-inline">{% translate "Manage federation" %}</span>
```
felixrindt, line 40: I think we should add `(experimental)` labels to buttons controlling these features.

ephios/api/templates/api/access_token_list.html
```diff
@@ -25,6 +25,9 @@ <h5 class="card-title">
                                 {{ token.application.name }}
                             </h5>
                             <p class="text-muted">
+                                {% if token.application.federatedhost_set.exists %}
+                                    {% translate "ephios instance" %}
+                                {% endif %}
```
felixrindt, line 30: I don't really like how the **plugin** federation spills over these specifics into the other app. Can we find another way to adapt the look of the application list and token list?

felixrindt: (Token list probably not even needs this hint)

ephios/plugins/federation/views/frontend.py
```diff
@@ -0,0 +1,169 @@
...
+    model = Event
+    template_name = "federation/event_detail.html"
+
+    def get_object(self, queryset=None):
+        obj = super().get_object(queryset)
+        try:
+            guest = self.request.session["guest"]
+            FederatedEventShare.objects.get(
+                event=obj,
+                shared_with__in=[FederatedGuest.objects.get(pk=guest)],
```
felixrindt, line 79: I get a `FederatedGuest.DoesNotExist` server error after recreating the connection and accessing the share again. Possibly because `guest` contains the old `FederatedGuest` primary key? Let me know if you need more help recreating this issue.

ephios/plugins/federation/templates/federation/external_event_list.html
```diff
@@ -0,0 +1,65 @@
...
+    <h1>{% translate "External events" %}</h1>
+    {% for event in events %}
+        <div class="list-group-item list-group-item-action p-0 d-flex flex-row">
+            <div class="m-0 py-2 d-flex flex-column flex-lg-row-reverse justify-content-around flex-grow-0">
+                <div class="ps-lg-2 d-flex flex-row flex-lg-column justify-content-center event-list-status-icon">
+                        <span class="text-muted far fa-user ps-2"></span>
+                </div>
+            </div>
+            <a class="list-group-item-action py-2 event-list-item-link"
+               href="{{ event.signup_url }}?referrer={{ SITE_URL }}">
```
felixrindt, line 18: why are you passing the SITE_URL here?

ephios/plugins/federation/templates/federation/event_detail.html
```diff
@@ -0,0 +1,9 @@
+{% extends "core/event_detail.html" %}
+{% load i18n %}
+
+{% block messages %}
+    <div class="alert alert-info" role="alert">
+        {% translate "You are currently viewing an event at another ephios instance." %}
+        <a href="{{ guest_url }}">{% translate "Back to your ephios instance" %}</a>
```
felixrindt, line 7: Maybe instead of using "ephios instance" here we use the organization names?

felixrindt: esp. to avoid use of the word instance

ephios/plugins/federation/views/api.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class SharedEventListView(ListAPIView):
+    serializer_class = SharedEventSerializer
+    permission_classes = [TokenHasScope]
+    required_scopes = []
+
+    def get_queryset(self):
+        try:
+            guest = self.request.auth.federatedguest_set.get()
```
felixrindt, line 37: can you add a comment that request.auth is the authtoken and federatedguest_set is the reverse accessor?

felixrindt, line 37: I was wondering why this doesn't break if I create multiple connections between the same instances, as FederatedGuest has a ForeignKey, not a OneToOne key. But of course access token differ between FederatedGuests. Still could be OneToOne couldn't it?

ephios/plugins/federation/forms.py
```diff
@@ -0,0 +1,102 @@
...
+        return _("Share event with other ephios instances")
+
+    def is_function_active(self):
+        return self.instance.pk and self.instance.shared_with.exists()
+
+
+class InviteCodeForm(forms.ModelForm):
+    class Meta:
+        model = InviteCode
+        fields = ["url"]
```
felixrindt, line 62: Make sure all the labels are translated, either in the model fields for model forms or right here as label on form fields. It should be "URL", not "Url" as a label.

felixrindt: Also I think some placeholder would be great to indicate that you must input `_("https://their-name.ephios.de")`

felixrindt (changes requested):

ephios/plugins/federation/templates/federation/invitecode_reveal.html
```diff
@@ -0,0 +1,50 @@
...
+        {% blocktranslate trimmed with url=invite.url %}
+            We created an invite code that allows the ephios instance at {{ url }} to accept your invitation to share events with them.
+            Please copy the URL or the code below and send it to the administrator of the other instance. They can then visit the URL or enter the code in their ephios instance to accept your invitation. It will expire in 48 hours.
+        {% endblocktranslate %}
+    </div>
+    <div class="mb-3">
+        <label for="url" class="form-label">Invite URL</label>
+            <div class="input-group">
+                <input id="url" class="form-control" type="text" aria-label="token"
+                       value="{{ invite.url|reverse_url:"federation:frontend_redeem_invite_code" }}?code={{ invite.get_share_string }}"
```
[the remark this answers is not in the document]
jeriox: but what if the user doesn't end the url with a slash? stuffing it together as a string doesn't respect that

ephios/plugins/federation/templates/federation/external_event_list.html
```diff
@@ -0,0 +1,65 @@
...
+                        </h5>
+                        <span class="w-100 text-muted text-break">
+                        {{ event.location }}
+                    </span>
+                    </div>
+                    <div class="grid-batch"><span
+                            class="badge bg-primary">{{ event.type }}</span>
+                    </div>
+                    <div class="grid-signup">
+                        <span class="fas fa-share"></span><span class="text-nowrap">{{ event.host }}</span>
```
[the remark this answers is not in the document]
jeriox: dunno anymore, I think it was even uglier without. Would be great if you could have a look

ephios/extra/templatetags/utils.py
```diff
@@ -55,3 +57,13 @@ def timedelta_in_hours(delta):
...
+
+
+@register.filter(name="urljoin")
+def _urljoin(url, path):
+    return urljoin(url, path)
+
+
+@register.filter(name="reverse_url")
+def reverse_url(absolute_url, urlname):
+    return urljoin(absolute_url, reverse(urlname))
```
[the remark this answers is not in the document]
jeriox: yeah I used urljoin in a previous version and thought I'd leave it there for future use
suggest a better name :D the regular tag for reversing is called url, so I wanted to go with something similar

ephios/api/templates/api/access_token_list.html
jeriox: > (Token list probably not even needs this hint)

I think it helps the user to understand where this is coming from

ephios/plugins/federation/views/frontend.py
jeriox: what do you mean by "recreating the connection"?

ephios/plugins/federation/templates/federation/external_event_list.html
jeriox: because the host needs to know where to send the user to complete the OAuth2 flow

ephios/plugins/federation/templates/federation/invitecode_reveal.html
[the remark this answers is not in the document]
felixrindt: True. I think the view should normalize the url anyway?! Forgot to check that when writing the comment though

ephios/plugins/federation/views/frontend.py
felixrindt: Removing the federation and then recreating the federation

ephios/plugins/federation/templates/federation/external_event_list.html
felixrindt: Ahh of course. I was wandering why it was popping up after having authorized the host already

ephios/plugins/federation/views/api.py
```diff
@@ -0,0 +1,125 @@
...
+                raise PermissionDenied from exc
+        else:
+            self._oauth_callback()
+            return redirect("federation:event_detail", pk=self.request.session.pop("event"))
+
+    def _get_authorization_url(self):
+        guest = (
+            FederatedGuest.objects.get(pk=self.request.session["guest"])
+            if "guest" in self.request.session.keys()
+            else FederatedGuest.objects.get(url=self.request.GET["referrer"])
```
felixrindt, line 58: these lines assume just on FederatedGuest per url and I was able to break that while testing around...

image

felixrindt (approved):

ephios/plugins/federation/views/api.py
felixrindt: No hard feelings but there are lots of places esp in this file where I think the code makes a lot of assumptions that might be broken with real users fiddeling around with this feature and I really don't want them to cause server errors.

felixrindt (changes requested):

ephios/plugins/federation/templates/federation/external_event_list.html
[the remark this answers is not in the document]
felixrindt: so as you are replacing a grid cell with auto width the text does not want to break. It was never meant to contain arbitrary length text so the css deals with it in an unhelpful way. 
A simple way to place the Host name somewhere would be to leave `grid-signup` entirely empty or put some "participation count unknown" thingy there and replace the `<h5/>` in `grid-title` with

```html
<div>
    <h5 class="mb-0 text-break d-inline-block">
        {{ event.title }}
    </h5>
    <small class="text-muted">
        <span class="fas fa-share"></span>
        <span class="">{{ event.host }}</span>
    </small>
</div>
```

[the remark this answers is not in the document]
felixrindt: so as you are replacing a grid cell with auto width the text does not want to break. It was never meant to contain arbitrary length text so the css deals with it in an unhelpful way. 
A simple way to place the Host name somewhere would be to leave `grid-signup` entirely empty or put some "participation count unknown" thingy there and replace the `<h5/>` in `grid-title` with

```html
<div>
    <h5 class="mb-0 text-break d-inline-block">
        {{ event.title }}
    </h5>
    <small class="text-muted">
        <span class="fas fa-share"></span>
        <span class="">{{ event.host }}</span>
    </small>
</div>
```

ephios/plugins/federation/views/frontend.py
```diff
@@ -0,0 +1,169 @@
...
+    def get_participant(self):
+        try:
+            return FederatedUser.objects.get(
+                pk=self.request.session["federated_user"]
+            ).as_participant()
+        except FederatedUser.DoesNotExist as e:
+            raise PermissionDenied from e
+
+
+class FederationSettingsView(TemplateView):
```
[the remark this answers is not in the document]
jeriox: views are now StaffRequired

ephios/plugins/federation/views/api.py
```diff
@@ -0,0 +1,125 @@
...
+class FederationOAuthView(View):
+    def get(self, request, *args, **kwargs):
+        if "code" not in request.GET.keys():
+            try:
+                return redirect(self._get_authorization_url())
+            except (KeyError, FederatedGuest.DoesNotExist) as exc:
+                raise PermissionDenied from exc
+        else:
+            self._oauth_callback()
+            return redirect("federation:event_detail", pk=self.request.session.pop("event"))
```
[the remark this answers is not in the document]
jeriox: good catch

ephios/core/templatetags/settings_extras.py
```diff
@@ -14,3 +14,10 @@ def available_management_settings_sections(request):
 @register.simple_tag
 def oidc_client_enabled():
     return settings.ENABLE_OIDC_CLIENT
+
+
+@register.simple_tag
+def organization_name():
+    from dynamic_preferences.registries import global_preferences_registry
+
+    return global_preferences_registry.manager()["general__organization_name"]
```
felixrindt, line 23: that might be something for `core/context.py`?!

ephios/api/templates/api/access_token_list.html
jeriox: we agreed that we will keep it but write a comment

ephios/plugins/federation/views/frontend.py
jeriox: I think that's impossible to prevent as we have no way of invalidating this session when the connection is removed. I added a `session.flush()` here, so the user would be presented with a permission denied, but on the next try everything would be set up correctly again

felixrindt: Of course the user has to give permission again - fine, but the `DoesNotExist` is preventable...

ephios/plugins/federation/signals.py
```diff
@@ -0,0 +1,71 @@
...
+
+@receiver(
+    management_settings_sections,
+    dispatch_uid="ephios.plugins.federation.signals.federation_settings_section",
+)
+def federation_settings_section(sender, request, **kwargs):
+    return (
+        [
+            {
+                "label": _("Federation") + " (beta)",
```
felixrindt, line 55: Can we replace "beta" with "experimental" (everywhere)? That's what I did in the serviceworker offline context.

ephios/plugins/federation/models.py
```diff
@@ -0,0 +1,136 @@
...
+            ).encode("ascii")
+        ).decode("ascii")
+
+
+class FederatedEventShare(models.Model):
+    event = models.ForeignKey(Event, on_delete=models.CASCADE)
+    shared_with = models.ManyToManyField(FederatedGuest)
+
+    def __str__(self):
+        return f"Federated event share for {self.event}"
```
felixrindt, line 65: translate

ephios/plugins/federation/models.py
```diff
@@ -0,0 +1,136 @@
...
+    first_name = models.CharField(_("first name"), max_length=254)
+    last_name = models.CharField(_("last name"), max_length=254)
+    date_of_birth = models.DateField(_("date of birth"))
+    phone = models.CharField(_("phone number"), max_length=254, blank=True)
+    qualifications = models.ManyToManyField(Qualification)
+    federated_instance = models.ForeignKey(FederatedGuest, on_delete=models.CASCADE)
+    created_at = models.DateTimeField(auto_now_add=True)
+
+    def __str__(self):
+        return f"Federated user {self.first_name} {self.last_name}"
```
felixrindt, line 79: translate

ephios/plugins/federation/signals.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@receiver(nav_link, dispatch_uid="ephios.plugins.federation.signals.nav_link")
+def add_nav_link(sender, request, **kwargs):
+    return [
+        {
+            "label": _("External events"),
+            "url": reverse("federation:external_event_list"),
+            "active": request.resolver_match and request.resolver_match.app_name == "federation",
+        }
```
felixrindt, line 23: should we only show this if we actually have a FederatedHost?!

ephios/plugins/federation/signals.py
```diff
@@ -0,0 +1,71 @@
...
+    ]
+
+
+@receiver(
+    participant_from_request,
+    dispatch_uid="ephios.plugins.federation.signals.guest_participant_from_request",
```
felixrindt, line 29:
```suggestion
    dispatch_uid="ephios.plugins.federation.signals.federated_participant_from_request",
```

felixrindt (changes requested):

ephios/plugins/federation/signals.py
felixrindt: (would need to be translated)

ephios/plugins/federation/models.py
```diff
@@ -0,0 +1,136 @@
...
+        return reverse("federation:event_detail", kwargs={"pk": event.pk})
+
+    @property
+    def icon(self):
+        # pylint: disable-next=consider-using-f-string
+        return mark_safe(
+            '<span class="fa fa-user-tag" data-bs-toggle="tooltip" data-bs-placement="left" title="{title}"></span>'.format(
+                title=_("Federated user")
+            )
+        )
```
felixrindt, line 123:
```suggestion
        title = _("Federated user")
        return mark_safe(
            f'<span class="fa fa-user-tag" data-bs-toggle="tooltip" data-bs-placement="left" title="{title}"></span>'
        )
```

tests/settings.py
```diff
@@ -1,3 +1,4 @@
 from ephios.settings import *  # no-qa
 
 LANGUAGE_CODE = "en"
+SITE_URL = "http://testserver"
```
felixrindt, line 4: shouldn't this overwrite `GET_SITE_URL`?

jeriox: I experimented a lot with how the live test server behaves esp. in conjunction with changes to the settings and it all seemed a bit messy. Not sure if this is still needed anyways, tests are still a big WIP ^^ (or rather WIB)

felixrindt: work in bahn? work in bitterness?

jeriox: work in backlog, but your options fit as well ^^

pyproject.toml
```diff
@@ -44,6 +44,9 @@ django-oauth-toolkit = "^2.3.0"
 mozilla-django-oidc = "^3.0.0"
 urllib3 = "^1.26.0,<2.0.0"  # pinned because of uberspace issues with urllib3 2.0.0
 pyyaml = "^6.0.1"
+lxml = "^4.9.3"
+beautifulsoup4 = "^4.12.2"
+css-inline = "^0.10.3"
```
felixrindt, line 49: where does this come from?

felixrindt (approved):

PR #976, opened 2023-06-16
Add File Logging

### LINKED ISSUE #349: Proper logging setup
We should provide proper logging configuration in `settings.py`. See https://docs.djangoproject.com/en/3.1/topics/logging/#configuring-logging

felixrindt:
* [x] set root log level for console logging to WARNING
* [x] create file logging with INFO level (incl rotation)

closes #349

### DIFF
--- a/ephios/core/plugins.py
+++ b/ephios/core/plugins.py
@@ -11,8 +11,6 @@ logger = logging.getLogger(__name__)
 
 # The plugin mechanics are heavily inspired by pretix (then licenced under Apache 2.0) - Check it out!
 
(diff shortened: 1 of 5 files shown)


ephios/core/plugins.py
```diff
@@ -11,8 +11,6 @@
 
 # The plugin mechanics are heavily inspired by pretix (then licenced under Apache 2.0) - Check it out!
 
-logger.info("Installed plugins: %s", ", ".join(settings.PLUGINS))
```
felixrindt, line 14: had to move that because Logging is only available after apps are loaded

jeriox (approved):

jeriox (approved):

PR #983, opened 2023-06-23
configurable env file location

### LINKED ISSUE #982: Read env-file from environment variable location
currently, this is a great way to get a dev setup running quickly:

https://github.com/ephios-dev/ephios/blob/9f37e9efb0e3395d0b6c599a0230449430e6cbdd/ephios/settings.py#L22

on production it would be great if you could specify where the env file is located.

felixrindt:
closes #982

### DIFF
--- a/ephios/settings.py
+++ b/ephios/settings.py
@@ -2,6 +2,7 @@ import copy
 import os
 from datetime import timedelta
 from email.utils import getaddresses
+from pathlib import Path
 
 import environ
 from cryptography.hazmat.primitives import serialization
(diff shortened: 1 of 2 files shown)


ephios/settings.py
```diff
@@ -14,12 +15,14 @@
 except ImportError:
     import importlib.metadata as importlib_metadata
 
-
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 env = environ.Env()
 # for syntax see https://django-environ.readthedocs.io/en/latest/
-environ.Env.read_env(env_file=os.path.join(BASE_DIR, ".env"))
+# read env file from ENV_PATH or fall back to a .evn file in the project root
```
jeriox, line 22: typo

ephios/settings.py
```diff
@@ -14,12 +15,14 @@
...
+env_path = env.str("ENV_PATH", default=os.path.join(BASE_DIR, ".env"))
+print(f"Loading ephios environment from {Path(env_path).absolute()}")
```
jeriox, line 24: sure you want to print stuff instead of logging it?

jeriox (approved):

ephios/settings.py
felixrindt: I will try again but it wasn't showing up using logging. Maybe that's because to configure logging, this file needs to be loaded first?

felixrindt: Ja well

PR #990, opened 2023-07-04
refactor template getter for shift configuration form rendering

jeriox:
all three implementations of `render_configuration_form()` were using the same line of code `form = form or self.get_configuration_form(*args, event=self.event, **kwargs)` and only a different template to render the form in. Unfortunately, the implementation in signup_beachguard was missing the `event=self.event` part which led to the form not being rendered. As the implementations were identical anyways, I decided to refactor the template getting into a separate method

### DIFF
--- a/ephios/core/signup/methods.py
+++ b/ephios/core/signup/methods.py
@@ -14,7 +14,6 @@ from django.core.exceptions import PermissionDenied, ValidationError
 from django.db import transaction
 from django.db.models import Q
 from django.shortcuts import redirect
-from django.template import Context, Template
 from django.template.loader import get_template
 from django.urls import reverse
 from django.utils import timezone
@@ -499,6 +498,7 @@ class BaseSignupMethod:
         return BaseDispositionParticipationForm
 
     configuration_form_class = BaseSignupMethodConfigurationForm
+    configuration_form_template_name = "core/signup_configuration_form.html"
 
     @property
     def signup_view_class(self):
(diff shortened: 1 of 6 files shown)


ephios/plugins/basesignup/signup/section_based.py
```diff
@@ -295,13 +295,9 @@ def _configure_participation(
...
 
-    def render_configuration_form(self, *args, form=None, **kwargs):
+    def get_configuration_form_template(self):
         """We overwrite the template to render the formset."""
-        form = form or self.get_configuration_form(*args, event=self.event, **kwargs)
-        template = get_template("basesignup/section_based/configuration_form.html").render(
-            {"form": form}
-        )
-        return template
+        return get_template("basesignup/section_based/configuration_form.html")
```
felixrindt, line 300: I wonder if we could be more close to how the django class based views work and just provide the class attribute `configuration_form_template_name` and just load from that name in `render_configuration_form`. That way you wouldn't need to overwrite a method, just the attribute - and you could reuse a template containing just

```jinja2
{% load crispy_forms_filters %}
{{ form|crispy }}
```

in both the base signup method and beachguard without having to repeat the basic implementation (as seen in https://github.com/ephios-dev/ephios/pull/990/files) or doing some complicated `super(SectionBasedSignupMethod, ...)` over there.

After all, signup methods are complicated already, so I think using established django petterns would fit here.

felixrindt: Maybe Django's new form Templates are also worth looking into...

jeriox: I think looking into form templates is something for later. I refactored to use an attribute, please re-review

ephios/core/signup/methods.py
```diff
@@ -768,11 +768,7 @@ def get_configuration_form(self, *args, **kwargs):
 
     def render_configuration_form(self, *args, form=None, **kwargs):
```
felixrindt, line 769: now do we even need this method? As I understand the django docs a simple `{{ form }}` should suffice?

felixrindt:
superseded by #996

Issue #853, opened 2023-01-02
StaticI18N catalog not loaded on some views


jeriox:
The shift edit views tries to load `/de.js` while other views like the instance settings are loading `/static/jsi18n/de/djangojs.js` correctly, despite originating from the same `<script type="text/javascript" src="{% statici18n LANGUAGE_CODE %}"></script>` in base.html

felixrindt:
I couldn't reproduce this

PR #998, opened 2023-07-11
Forms Template/JS Refactoring

felixrindt:
I intend to merge this with rebase (no squash).

* refactor PluginForms to use form templates instead of FormHelper
* update jQuery to 3.7.0
* [reduce flicker in formset handler](https://github.com/ephios-dev/ephios/pull/998/commits/6b176efd6911348b84bca02a65426a695f744695) by 
    * hiding the new form before it is added to the dom when using animation, therefore no flicker when clicking the add button
    * creating a new event type to `handleForms` before the slideDown animation. Before, the animation would compute before `handleForms` was called and therefore use the wrong html (e.g. the vanilla select-multiple instead of select2 which is smaller) leading to jumps at the end of the animation


ephios/plugins/simpleresource/forms.py
```diff
@@ -17,9 +16,8 @@ class Meta:
         }
 
     def __init__(self, *args, shift, **kwargs):
+        kwargs.setdefault("prefix", "simple_resource")
         self.shift = shift
-        self.helper = FormHelper()
-        self.helper.include_media = False
```
felixrindt, line 22: @jeriox why was this here? Didn't find another use of it...

felixrindt: was used to fix https://github.com/ephios-dev/ephios/issues/853 but I don't understand how

jeriox: jsi18n offers to handle the loading of the translation catalogue for you by using the django forms media interface. This includes the JS file, but from a path that doesn't match to what we are doing. In the base, we are importing the catalogue ourselves, so it works everywhere else. Using the form helper and the template tag here meant that crispy also injected the form media, which meant the erroneous JS path. Setting this to false prevents crispy from doing that and everything is fine. So as long as you don't use the crispy tag all should be fine

jeriox (approved):

ephios/plugins/simpleresource/forms.py
felixrindt: Alright, so as we are not using the tag here, this is fine. Sounds like another reason to rethink our form rendering though. 🙈

PR #999, opened 2023-07-13
I18n housekeeping

### LINKED ISSUE #830: Check for untranslated strings
Right now, there are a lot of strings that either aren't using the translate tag or are missing a translation in weblate. This impacts the UX and the understanding significantly. We should check for those strings.

felixrindt:
Oh boy, I didn't know you had to run `makemessages` with the `djangojs` domain?!

This I think for the most part fixes #830

### DIFF
--- a/ephios/core/signup/participants.py
+++ b/ephios/core/signup/participants.py
@@ -114,6 +114,9 @@ class PlaceholderParticipant(AbstractParticipant):
 
     @property
     def icon(self):
+        title = _(
+            "Placeholder"
+        )  # not getting marked for translation when inside f-string for some reason
         return mark_safe(
-            f'<span class="fa fa-user-tag" data-toggle="tooltip" data-placement="left" title="{_("Placeholder")}"></span>'
+            f'<span class="fa fa-user-tag" data-toggle="tooltip" data-placement="left" title="{title}"></span>'
         )
--- a/ephios/templates/base.html
+++ b/ephios/templates/base.html
@@ -50,13 +50,12 @@
         {% block css %}{% endblock %}
     {% endcompress %}
 
-    {# as of 2021-06-14, the statici18n tag misbehaves with compression when 500 errors occur #}
-    <script type="text/javascript" src="{% statici18n LANGUAGE_CODE %}"></script>
     {% compress js %}
         <script type="text/javascript" src="{% static "jquery/js/jquery-3.7.0.min.js" %}"></script>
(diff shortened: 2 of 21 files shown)


ephios/core/signup/participants.py
```diff
@@ -114,6 +114,9 @@ def reverse_event_detail(self, event):
 
     @property
     def icon(self):
+        title = _(
+            "Placeholder"
+        )  # not getting marked for translation when inside f-string for some reason
```
jeriox, line 119: yeah I think that's known and django suggests to use the old syntax

ephios/templates/base.html
```diff
@@ -50,13 +50,12 @@
         {% block css %}{% endblock %}
     {% endcompress %}
 
-    {# as of 2021-06-14, the statici18n tag misbehaves with compression when 500 errors occur #}
```
jeriox, line 53: did you find out why it happened? did it stop happening now?

ephios/urls.py
```diff
@@ -28,7 +27,6 @@
     path("", include("ephios.core.urls")),
     path("api/oauth/", include("ephios.api.access.oauth2_urls", namespace="oauth2_provider")),
     path("api/", include("ephios.api.urls")),
-    path("jsi18n.js", JavaScriptCatalog.as_view(packages=["recurrence"]), name="jsi18n"),
```
jeriox, line 31: where does this come from now instead?

ephios/templates/base.html
felixrindt: so I ran a test with DEBUG=False and its all fine and generating a response with our 500.html template. As that template is not based on the `base.html` template anyway, there is no reason to fail now. (although it looks dull :sweat_smile: )

ephios/urls.py
felixrindt: Actually, it was never used! There are two options to deliver the i18n js:

1. Djangos built-in solution is to use the `JavaScriptCatalog`, include it in the urls as above **and** load it in the template (which we never did, because we use option 2).
2. `django-statici18n`, which collects the catalog for every language into a folder so it does not have to be generated on the fly for every (non-cached) request.

With option 2, we can include the catalog in compression.

ephios/core/signup/participants.py
felixrindt: seems to be a limitation in gettext

jeriox (approved):

PR #1000, opened 2023-07-14
PWA more offline availability

felixrindt:
This updates the PWA views to class based views and adds a few more fields. Then there are some random fixes and touchups. Biggest change (and the most debatable) is the new serviceworker that caches sites you visit on the go.

### DIFF
--- a/ephios/static/ephios/js/main.js
+++ b/ephios/static/ephios/js/main.js
@@ -69,7 +69,6 @@ $(document).ready(function () {
         $('#unloading-spinner').addClass("d-none");
     });
 
-    // if loaded offline, show a warning instead of maybe cached messages
     if ($("body").data("pwa-network") === "offline") {
         // disable all forms and post buttons
         $("form").find("input, [type='submit'], select, textarea").prop("disabled", true);
--- a/ephios/static/ephios/js/main.js
+++ b/ephios/static/ephios/js/main.js
@@ -78,8 +78,8 @@ $(document).ready(function () {
         time = new Date(time).toLocaleString(document.querySelector("html").getAttribute('lang'));
         message(
             "warning",
-            gettext("You are currently offline and seeing the page as it was on {time}.").replace("{time}", time),
-            0
+            gettext("You are currently offline and seeing the content as it was on {time}.").replace("{time}", time)
+            + " (" + gettext("experimental") + ")", 0
         );
     }
 
--- a/ephios/static/ephios/js/main.js
+++ b/ephios/static/ephios/js/main.js
@@ -83,7 +83,7 @@ $(document).ready(function () {
         );
     }
 
-    // In itialize the service worker
+    // Initialize the service worker
     if ('serviceWorker' in navigator) {
         navigator.serviceWorker.register('/serviceworker.js', {
             scope: '/'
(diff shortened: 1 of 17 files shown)


ephios/static/ephios/js/serviceworker.js
```diff
@@ -19,40 +8,70 @@ const filesToCache = [
...
-// Serve from Cache
 self.addEventListener("fetch", event => {
     event.respondWith(
-        caches.match(event.request)
-            .then(response => {
-                return response || fetch(event.request);
-            })
-            .catch(() => {
-                return caches.match('/offline/');
-            })
```
felixrindt, line 56: Problem might actually be in the cache management. See how here I'm looking at *all* caches while in the new version only in the cache newly selected.

ephios/static/ephios/js/serviceworker.js
```diff
@@ -19,40 +8,86 @@ const filesToCache = [
...
-    )
-});
+    const isStatic = new URL(event.request.url).pathname.startsWith("/static/");
```
felixrindt, line 87: could use another way to descriminate between caching strategies:
image
(from https://developer.chrome.com/docs/workbox/caching-strategies-overview/)

ephios/static/ephios/js/serviceworker.js
```diff
@@ -19,40 +8,86 @@ const filesToCache = [
...
+    if (isStatic) {
+        event.respondWith(cacheThenNetwork(event));
+    } else {
+        event.respondWith(networkThenCacheOrOffline(event));
+    }
```
felixrindt, line 92: This is choosing different caching strategies for different paths, a pattern implemented in googles 'workbox' library for building serviceworkers in a node environment. It's really hard to understand end to end though, and requires you to not only learn serviceworker caching, but workbox as well.

ephios/static/ephios/js/serviceworker.js
```diff
@@ -19,40 +8,86 @@ const filesToCache = [
...
                     .filter(cacheName => (cacheName !== staticCacheName))
                     .map(cacheName => caches.delete(cacheName))
             );
         })
     );
 });
 
-// Serve from Cache
+async function markResponseAsOffline(response) {
+    let content = await response.body.getReader().read().then(r => r.value);
```
felixrindt, line 35: I think this line is a great demonstration of javascripts async API.
`await` obvs awaits a Promise to be fulfilled, but look how this is actually a chained promise because of the `then` at the end.

jeriox (approved):

jeriox:
When trying this with the offline simulator in Chrome Desktop, the CSS of the offline page is not rendered. Service worker has been freshly installed.
grafik
grafik

ephios/static/ephios/js/main.js
```diff
@@ -65,19 +68,33 @@ $(document).ready(function () {
...
-if ('serviceWorker' in navigator) {
-    navigator.serviceWorker.register('/serviceworker.js', {
-        scope: '/'
-    }).then(function (registration) {
-        console.log('django-pwa: ServiceWorker registration successful with scope: ', registration.scope);
-    }, function (err) {
-        console.log('django-pwa: ServiceWorker registration failed: ', err);
-    });
-}
+    // In itialize the service worker
```
felixrindt, line 86:
```suggestion
    // Initialize the service worker
```

PR #1011, opened 2023-08-02
add missing dependency for schema generation view

felixrindt:


pyproject.toml
```diff
@@ -43,6 +43,7 @@ redis = {version = "^4.3.4", optional = true}
 django-oauth-toolkit = "^2.3.0"
 mozilla-django-oidc = "^3.0.0"
 urllib3 = "^1.26.0,<2.0.0"  # pinned because of poetry issues with urllib3 2.0.0
+pyyaml = "^6.0.1"
```
jeriox, line 46: why is this required? is this just a missing dependency in another package?

ephios/settings.py
```diff
@@ -32,7 +32,7 @@
 DEBUG = env.bool("DEBUG")
 ALLOWED_HOSTS = env.list("ALLOWED_HOSTS")
 
-EPHIOS_VERSION = "v" + importlib_metadata.version("ephios")
+EPHIOS_VERSION = importlib_metadata.version("ephios")
```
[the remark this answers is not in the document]
felixrindt: Yes. I thought about adding it in the footer template but I think it's also fine without isn't it?

pyproject.toml
felixrindt: It's required for the SchemaView of drf. See the documentation there...

felixrindt: image
https://www.django-rest-framework.org/api-guide/schemas/

ephios/settings.py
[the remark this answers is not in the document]
jeriox: i think i like it shlighty more with the v, but no hard feelings

pyproject.toml
jeriox: wow. still think that it _should_ be their job to specify that ^^

jeriox (approved):

ephios/settings.py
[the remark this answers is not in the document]
felixrindt: You know I was thinking reading the footer flows better with the v there. Otherwise your mind has to add the "version" when reading. I'll add it back in...

pyproject.toml
felixrindt: well its really an optional feature but yeah

PR #1016, opened 2023-08-04
Fix workinghour mixin

felixrindt:


ephios/core/views/workinghour.py
```diff
@@ -23,15 +24,15 @@
...
+        """Return whether the current request user can grant working hours to the target user (which might be themselves)"""
+        if self.request.user.is_anonymous:
+            return False
+        can_grant_for_groups = get_objects_for_user(
             self.request.user, "decide_workinghours_for_group", klass=Group
-        ).values_list("id", flat=True)
-        self.can_grant = (
-            self._get_target_user().groups.filter(id__in=can_grant_for_group_ids).exists()
         )
-        return result
```
felixrindt, line 34: actually, `setup` doesn't return anything

ephios/core/views/workinghour.py
```diff
@@ -23,15 +24,15 @@
...
+        return self._get_target_user().groups.filter(id__in=can_grant_for_groups).exists()
```
felixrindt, line 35: Problem was that when `request.user` is anonoumous, the `get_objects_for_user` was failing because we don't use guardians anonymous user feature. But at the view's setup stage, the redirect to login did not occur yet. I fixed it first for the detail view by making it a property only computed after the view's disposition stage in which an anonymous user might be redirected to login. But that didn't help with the views that require users to `can_grant` to have permission in the first place. That's when I added the 

```
if self.request.user.is_anonymous:
    return False
```

which I think would be enough in the first place, though I do like it more this way either way.

ephios/core/views/workinghour.py
```diff
@@ -22,19 +23,22 @@
...
 
     def _get_target_user(self):
-        """Return the working hour user"""
+        """Return the user whose working hours are being managed."""
         return get_object_or_404(UserProfile, pk=self.kwargs["pk"])
```
felixrindt, line 42: Problem was that when `request.user` is anonoumous, the `get_objects_for_user` was failing because we don't use guardians anonymous user feature. But at the view's setup stage, the redirect to login did not occur yet. I fixed it first for the detail view by making it a property only computed after the view's disposition stage in which an anonymous user might be redirected to login. But that didn't help with the views that require users to `can_grant` to have permission in the first place. That's when I added the 

```
if self.request.user.is_anonymous:
    return False
```

which I think would be enough in the first place, though I do like it more this way either way. With the new semantics I also changed the class name.

jeriox (approved):

PR #1018, opened 2023-08-06
move test event to the far future

felixrindt:


jeriox (approved):
what about the tests? I guess they will also fail at some point (signup) because the event is in the past?

PR #1020, opened 2023-08-06
fix custom participant notification sent twice

### LINKED ISSUE #959: Custom participant notification is sent twice
When a user is both participating and resposible for an event, they will receive custom notification texts twice, once for each role. Instead, they should only receive one notification.

olewun:
Hard work for Franz.

Closes #959


ephios/core/services/notifications/types.py
```diff
@@ -452,11 +452,16 @@ class CustomEventParticipantNotification(AbstractNotificationHandler):
     @classmethod
     def send(cls, event: Event, content: str):
         participants = set()
+        responsible_users = get_users_with_perms(
+            event, with_superusers=False, only_with_perms_in=["change_event"]
+        )
         for shift in event.shifts.all():
             participants.update(shift.get_participants())
         notifications = []
         for participant in participants:
```
jeriox, line 461: instead of using the if statement below, you could also use the [queryset difference](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#difference) to make it a bit more readable/understandable what happens here. But it's also fine the way it is

jeriox (approved):
thanks for your contribution!

ephios/core/services/notifications/types.py
felixrindt: How would one do that? How would the difference look here? I think that would be far more complex...

jeriox: `participants.difference(responsible_users)`?

felixrindt: I think you are mistaken, participants is not a queryset. Participants aren't even model instances.

jeriox: ah yeah

PR #1021, opened 2023-08-06
allow case insensitive login for postgres

jeriox:


felixrindt (changes requested):
please add tests, e.g. that

* a manager can't create another user with the same email with uppercase letters
* login works with uppercase email stored in lower case
* ...

felixrindt (approved):

PR #1028, opened 2023-08-14
revert poetry-publish to v1.16

jeriox:


felixrindt (approved):
Can you explain?

jeriox:
the latest proposal from the author to use a container didn't work, and everyone else in the issue reported that downgrading to v1.16 solved the issue

PR #1036, opened 2023-08-24
Forbid deleting the last staff user

felixrindt:
depends on https://github.com/ephios-dev/ephios/pull/1037 so rebase main onto that first

- [x] protect actions that lead to no active user having management permissions (removing users from groups etc)
- [x] add more tests

### DIFF
--- a/ephios/core/forms/users.py
+++ b/ephios/core/forms/users.py
@@ -25,7 +25,7 @@ from ephios.core.services.notifications.types import enabled_notification_types
 from ephios.core.signals import register_group_permission_fields
 from ephios.core.widgets import MultiUserProfileWidget
 from ephios.extra.crispy import AbortLink
-from ephios.extra.permissions import PermissionField, PermissionFormMixin
+from ephios.extra.permissions import PermissionField, PermissionFormMixin, get_groups_with_perms
 from ephios.extra.widgets import CustomDateInput
 from ephios.modellogging.log import add_log_recorder
 from ephios.modellogging.recorders import DerivedFieldsLogRecorder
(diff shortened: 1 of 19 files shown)


jeriox (approved):
do we want to implement similar checks for at least one group with the "can manage ephios instance" permission? AFAIS you could loose that permission and even staff users wouldn't be able to recover it

felixrindt:
> do we want to implement similar checks for at least one group with the "can manage ephios instance" permission? AFAIS you could loose that permission and even staff users wouldn't be able to recover it

I think that's a very good idea. UX wise we could also just disable the delete button instead of showing the error only after people hit delete. What do you think?

jeriox:
> > do we want to implement similar checks for at least one group with the "can manage ephios instance" permission? AFAIS you could loose that permission and even staff users wouldn't be able to recover it
> 
> I think that's a very good idea. UX wise we could also just disable the delete button instead of showing the error only after people hit delete. What do you think?

yes that would be a bit smoother, but I could also live with the version you implemented here

ephios/extra/permissions.py
```diff
@@ -22,29 +26,70 @@ def _wrapped_view(request, *args, **kwargs):
...
+    return qs.distinct()
+
 
-    group_rel_name = group_model.group.field.related_query_name()
+def get_groups_with_all_permissions_in(permission_list: Iterable[str]):
+    permissions = get_permissions_from_qualified_names(permission_list)
+    qs = Group.objects.all()
+    for permission in permissions:
+        qs = qs.filter(permissions=permission)
+    return qs
```
felixrindt, line 84: Remove

jeriox (approved):

ephios/core/forms/users.py
```diff
@@ -200,14 +218,22 @@ def save(self, commit=True):
...
-class UserProfileForm(ModelForm):
+class UserProfileForm(PermissionFormMixin, ModelForm):
     groups = ModelMultipleChoiceField(
         label=_("Groups"),
         queryset=Group.objects.all(),
         widget=Select2MultipleWidget,
         required=False,
         disabled=True,  # explicitly enable for users with `change_group` permission
     )
+    is_staff = PermissionField(
```
jeriox, line 229: does this still set the actual is_staff flag? couldn't find that in the code

jeriox (approved):

ephios/core/forms/users.py
felixrindt: well `PermissionField` is just a BooleanField on a ModelForm so it does set the flag, yes

felixrindt: could rename it to `PermissionBooleanField` to make that more clear as well as emphasize the it's either one or off, not a select.

jeriox: > well `PermissionField` is just a BooleanField on a ModelForm so it does set the flag, yes

ah well. we just didn't use it that way before, because it only operated as an additional field that doesn't exist on the model itself. just a bit intransparent that it updates the flag as well as the permissions

felixrindt: what docstring? of `PermissionField` or the Form I use it in?

jeriox: PermissionField

PR #1044, opened 2023-08-30
Select2 bootstrap 5 theme fix

felixrindt:
This integrates a different theme pack into our saas compilation so the custom colors are right.


felixrindt:
Looks great now, apart from the bottom border being cut off...

image

image

jeriox (approved):

PR #1048, opened 2023-08-31
Fix month query for some mariadb

felixrindt:
A user reported not seeing their shifts in the calendar view. We debugged it together - turns out their mysql server was not giving back the shifts with the `__month` query. This PR replaces it with a datetime query which did resolve the issue for them.

### DIFF
--- a/ephios/core/views/event.py
+++ b/ephios/core/views/event.py
@@ -280,13 +280,20 @@ class EventListView(LoginRequiredMixin, ListView):
             .prefetch_related("participations")
         )
         year, month = self.filter_form.get_calendar_month()
-        shifts = shifts.filter(start_time__year=year, start_time__month=month)
+        prevyear, prevmonth = _prevmonth(year, month)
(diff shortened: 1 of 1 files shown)


ephios/core/views/event.py
```diff
@@ -280,13 +280,19 @@ def _get_calendar_context(self):
             .prefetch_related("participations")
         )
         year, month = self.filter_form.get_calendar_month()
-        shifts = shifts.filter(start_time__year=year, start_time__month=month)
+        prevyear, prevmonth = _prevmonth(year, month)
+        nextyear, nextmonth = _nextmonth(year, month)
+
+        from_time = datetime.min.replace(year=year, month=month, tzinfo=get_current_timezone())
```
jeriox, line 286: pls add a comment explaining that we are not using the `__month` query because of mysql as we might stumble upon that in the future

jeriox (approved):

PR #1058, opened 2023-09-06
Add HSTS seconds env var

### LINKED ISSUE #1056: Strict Transport Security (HSTS) duration ist too short
**Describe the bug**
Using tools like Mozilla Observatory to check the security of an Ephios instance it comes the error that HSTS is implemented incorrectly, because the specified duration is too short (max-age parameter). This is defined in the settings.py with 360 secounds. Normally, a time indication of half to one year is recommended. 

**To Reproduce**
- Got to  https://observatory.mozilla.org/ or https://www.ssllabs.com/ssltest/
- Enter the UrL of the Ephios instance and start the test. 
- Lock at HTTP Strict Transport Security 

**Expected behavior**
The variable SECURE_HSTS_SECONDS in settings.py must be set to a higher value, at least 15552000 seconds (better 31536000), by default. 

**Environment**
I'm using Ubuntu 22.04 server edition with Apache2 as reverse Proxy.

felixrindt:
closes #1056


ephios/settings.py
```diff
@@ -46,8 +46,8 @@
     X_FRAME_OPTIONS = "DENY"
     SECURE_CONTENT_TYPE_NOSNIFF = True
     SECURE_SSL_REDIRECT = True
-    SECURE_HSTS_SECONDS = 3600
-    SECURE_HSTS_INCLUDE_SUBDOMAINS = True
+    SECURE_HSTS_SECONDS = env.int("SECURE_HSTS_SECONDS", default=3600 * 24 * 366)
```
jeriox, line 49: I would prefer having an additional ENABLE_HSTS flag. That way, if HSTS is enabled, the default for the duration is sensible. On the other hand, defaulting the flag to false means that HSTS won't activate without the user understanding the consequences of it being set to one year/the server configured with improper SSL

pov91:
I think the easiest way is to just increment the variable to 1 year in settings.py. Of course it is possible to make this configurable in the .env file. Then the suggested setting in example files or on the setup page should be such that HSTS is activated with a duration of one year.

ephios/settings.py
felixrindt: We could just default to hsts 3600 seconds and put a section in the deployment guide on ssl hardening

felixrindt: We could just default to hsts 3600 seconds and put a section in the deployment guide on ssl hardening

pov91: Why do you want 3600 seconds as default if there is significantly more recommended everywhere?

felixrindt: We want a long hsts to be configured explicitly after people verified that their setup is working. Otherwise they might lock browsers out (because of how HSTS works).

Thats what recommended in the Django documentation as well.

felixrindt: Don't get me wrong, we want admins to have a secure setup, and that includes a long HSTS, but copying numbers, shell commands and config files is not the full story here.

jeriox (approved):

PR #1059, opened 2023-09-06
Do not allow ambigous datetimes in shift form

felixrindt:


ephios/core/forms/events.py
```diff
@@ -210,20 +211,25 @@ def __init__(self, *args, **kwargs):
...
-            cleaned_data["start_time"] = make_aware(
-                datetime.combine(cleaned_data["date"], cleaned_data["start_time"])
-            )
-            cleaned_data["end_time"] = make_aware(
-                datetime.combine(self.cleaned_data["date"], cleaned_data["end_time"])
-            )
-            if self.cleaned_data["end_time"] <= self.cleaned_data["start_time"]:
-                cleaned_data["end_time"] = cleaned_data["end_time"] + timedelta(days=1)
+        if {"meeting_time", "start_time"} <= set(cleaned_data.keys()):
             if not cleaned_data["meeting_time"] <= cleaned_data["start_time"]:
```
jeriox, line 233: is it a good idea that we just shift the start time by one day but not the meeting time? (can also be a part of rethinking shifts with respect to >24h to get the bugfix out quickly)

jeriox (approved):

PR #1064, opened 2023-09-12
Add docker deployment

felixrindt:
I added a Dockerfile and I'm trying to figure out how to deploy it using docker-compose.
I also changed the `settings.py` to reduce the amount of variables that need to be configured, streamlining the deployment for manual deployments as well.

## TODO

* [x] fix CSRF_TRUSED_ORIGINS
* [x] update manual deployment documentation for new env variables
* [x] write docker documentation
* [x] update our deployments, tests and docs to use the new variables
* [x] container health check
* [x] resolve warnings in tests
* [x] github workflow
* [x] integrate health checks


ephios/settings.py
```diff
@@ -16,44 +18,59 @@
...
+    else:
+        chars = string.ascii_letters + string.digits + string.punctuation
+        SECRET_KEY = get_random_string(50, chars)
+        with open(SECRET_FILE, "w") as f:
+            os.chmod(SECRET_FILE, 0o600)
+            try:
+                os.chown(SECRET_FILE, os.getuid(), os.getgid())
+            except AttributeError:
+                pass  # os.chown is not available on Windows
+            f.write(SECRET_KEY)
```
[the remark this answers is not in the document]
jeriox: I don't really like all of the setup stuff here. How about we create a initialize_project managemant command that sets up the directories, creates vapid keys, creates secret key etc.

ephios/settings.py
```diff
@@ -16,44 +18,59 @@
...
-ALLOWED_HOSTS = env.list("ALLOWED_HOSTS")
+if DEBUG:
+    DATA_DIR = env.str("DATA_DIR", os.path.join(BASE_DIR, "data"))
+else:
+    # DATA_DIR must be set explicitly in production
+    DATA_DIR = env.str("DATA_DIR")
+
+LOG_DIR = env.str("LOG_DIR", os.path.join(DATA_DIR, "logs"))
+MEDIA_ROOT = env.str("MEDIA_ROOT", os.path.join(DATA_DIR, "media"))
+STATIC_ROOT = env.str("STATIC_ROOT", os.path.join(DATA_DIR, "static"))
```
jeriox, line 41: I think we should keep to option to specify the static directory as people may want to put that somewhere under /var/www or similiar

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    rm -rf /var/lib/apt/lists/* \
+RUN dpkg-reconfigure locales && \
+	locale-gen C.UTF-8 && \
+	/usr/sbin/update-locale LANG=C.UTF-8 \
+RUN mkdir -p /data/static/
+RUN pip install "poetry==$POETRY_VERSION" gunicorn
+
+COPY pyproject.toml /app/pyproject.toml
+COPY poetry.lock /app/poetry.lock
+RUN poetry install -E pgsql -E redis -E mysql
```
jeriox, line 24: maybe install all extras instead?

deployment/docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+    command: gunicorn
+    volumes:
+      - ephios_data_volume:/var/ephios/data/
+    depends_on:
+      - ephios_postgres
+    env_file:
+      - .env
+    networks:
+      - intern
+    hostname: django.ephios.de
```
jeriox, line 16: probably no FQDN under our domain for public templates

ephios/settings.py
felixrindt: we are keeping it! You can still configure STATIC_ROOT!

ephios/settings.py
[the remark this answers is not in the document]
felixrindt: I think it might get fiddly as well. For the management command to run the settings must load and the settings need *some* value for the secret key

Dockerfile
felixrindt: yeah, can do `--all-extras --without=dev`

ephios/settings.py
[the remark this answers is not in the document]
felixrindt: I did move most of the code outside. I think it reads cleaner now.

docs/admin/deployment/docker/index.rst
```diff
@@ -0,0 +1,55 @@
...
+Docker Image
+------------
+
+We automatically build a Docker image for every release.
+It uses gunicorn as a WSGI-server and still requires a
+database, redis and some ssl-terminating webserver/proxy
+to be set up. Also, static files need to be served by
+a webserver from a common volume.
+
+The image is available on the Github Container Registry.
```
felixrindt, line 13: Put the link here

.github/workflows/docker.yml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'main'
+    tags:
+      - 'v*'
+  pull_request:
+    branches:
+      - 'main'
```
felixrindt, line 12: Kann weg

.github/workflows/docker.yml
```diff
@@ -0,0 +1,43 @@
+
+name: docker
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'main'
+    tags:
+      - 'v*'
```
felixrindt, line 9: Does this only build on tags? We should also provide a "dev" image.

ephios/settings.py
```diff
@@ -11,49 +11,62 @@
...
+    "PUBLIC_DIR": PUBLIC_DIR,
+    "STATIC_ROOT": STATIC_ROOT,
+    "PRIVATE_DIR": PRIVATE_DIR,
+    "LOG_DIR": LOG_DIR,
+    "MEDIA_ROOT": MEDIA_ROOT,
+}
+
+for path in DIRECTORIES.values():
+    if not os.path.exists(path):
+        os.makedirs(path, exist_ok=True)
```
felixrindt, line 53: move to management command + check

ephios/settings.py
```diff
@@ -374,28 +382,33 @@
...
-    }
-
+VAPID_PRIVATE_KEY_PATH = env.str(
+    "VAPID_PRIVATE_KEY_PATH", os.path.join(PRIVATE_DIR, "vapid_key.pem")
+)
+if not os.path.exists(VAPID_PRIVATE_KEY_PATH):
+    vapid = Vapid()
+    vapid.generate_keys()
+    vapid.save_key(VAPID_PRIVATE_KEY_PATH)
+    vapid.save_public_key(VAPID_PRIVATE_KEY_PATH + ".pub")
```
felixrindt, line 392: move to check and management command

ephios/core/views/healthcheck.py
```diff
@@ -0,0 +1,26 @@
...
+    def get(self, request, *args, **kwargs):
+        okays = []
+        not_okays = []
+
+        for check, status, message in run_healthchecks():
+            text = f"{check.name}: {message}"
+            if status == HealthCheckStatus.OK:
+                okays.append(text)
+            else:
+                not_okays.append(text)
```
felixrindt, line 17: only errors should be not-okay

deployment/compose/docker-compose.yaml
```diff
@@ -0,0 +1,49 @@
+services:
+  ephios:
+    build: ../../
```
jeriox, line 3: should be image from registry?

jeriox (approved):

PR #1080, opened 2023-09-24
save language preference for user

### LINKED ISSUE #1076: Respect user language
As a user, I want to use ephios in my own language. This includes the web interface (language could be determined by browser) and notifications (language needs to be set before)

jeriox:
closes #1076


ephios/core/services/notifications/backends.py
```diff
@@ -32,6 +33,8 @@ def send_all_notifications():
         for notification in Notification.objects.filter(failed=False):
             if backend.can_send(notification) and backend.user_prefers_sending(notification):
                 try:
+                    if notification.user:
+                        translation.activate(notification.user.language)
                     backend.send(notification)
```
felixrindt, line 38: I thought this would be a context manager...

ephios/extra/middleware.py
```diff
@@ -0,0 +1,13 @@
...
+
+class EphiosLocaleMiddleware(LocaleMiddleware):
+    def process_response(self, request, response):
+        response = super().process_response(request, response)
+        if settings.LANGUAGE_COOKIE_NAME not in request.COOKIES:
+            try:
+                response.set_cookie(settings.LANGUAGE_COOKIE_NAME, request.user.language)
+            except (KeyError, AttributeError):
+                pass
+        return response
```
felixrindt, line 13: This doesn't set the db field to the browser preference?!

felixrindt (commented):
Will need to have a look with my computer too

ephios/extra/middleware.py
jeriox: no, this stores the saved user preference into the cookie to be used in the session

ephios/core/services/notifications/backends.py
jeriox: the django docs don't say that this is possible, they just set the language back to the previous value if needed, but I figured that wouldn't be necessary here

ephios/core/forms/users.py
```diff
@@ -464,3 +464,9 @@ def __init__(self, *args, **kwargs):
 
     def update_preferences(self):
         self.user.preferences["notifications__notifications"] = self.cleaned_data
+
+
+class UserOwnDataForm(ModelForm):
+    class Meta:
+        model = UserProfile
+        fields = ["language"]
```
felixrindt, line 472: I'd prefer if the label would be "Interface Language" as it does not change the language of user provided values (e.g. event description).

felixrindt (approved):

ephios/core/forms/users.py
jeriox: is now preferred language

Issue #146, opened 2020-09-26
Support Name Titles


felixrindt:
**Is your feature request related to a problem? Please describe.**

My name is Dr. Lilly Mayer, so I put "Lilly" as my given name and "Dr. Mayer" as my last name. 
In some places, my name is displayed as "Lilly Dr. Mayer", which I find strange.

**Describe the solution you'd like**
I want my name to be displayed as "Dr. Lilly Mayer".
I want to input my title separately from my last name.

**Describe alternatives you've considered**

* let user choose a display string. That would support nicknames as well, but could differ a lot from first/last names shown in other places.
* support multiple naming schemes like pretix does.

jeriox:
could you link to/explain the "multiple naming schemes" thing?
+1 in general.
We could also think of Anreden at this point.

felixrindt:
pretix blog post about name schemes: https://pretix.eu/about/en/blog/20181109-release-220/
A simple Title field should be fine though

Could we avoid Anreden and the complexity of gender by sticking to "Hello {user}"?

jeriox:
> By now and after having read https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/ multiple times I think something like "display name" and "full name" might be what I want

that would map to Felix being your display name and Felix Rindt being your full name?

felixrindt:
> > By now and after having read https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/ multiple times I think something like "display name" and "full name" might be what I want
> 
> that would map to Felix being your display name and Felix Rindt being your full name?

depends really on what people prefer or what seems to be the standard in their organization. Display name is the one used to display the user in all kinds of views and email templates.

The full_name might be used for search, exporting certificates, etc. (I don't think we have much use for it now).

We could even have a seperate greeting (name), if people want to be greeted with a different name (e.g. in emails) than shown to other users.

PR #1088, opened 2023-10-01
customizable OIDC

jeriox:
- [x] CRUD views
- [x] OIDC discovery
- [x] setting to disable regular login form
- [x] tests
- [x] take groups from OIDC -> #1105


ephios/core/views/auth.py
```diff
@@ -0,0 +1,51 @@
...
+from django.utils.translation import gettext as _
+from django.views.generic import RedirectView
+from oauthlib.oauth2 import WebApplicationClient
+from requests_oauthlib import OAuth2Session
+
+from ephios.core.models.users import EphiosOIDCClient
+
+
+class OAuthRequestView(RedirectView):
+    def get_redirect_url(self, *args, **kwargs):
```
felixrindt, line 15: This screams of key/doesnotexist errors

ephios/extra/auth.py
```diff
@@ -18,3 +37,32 @@ def update_user(self, user, claims):
...
+            if "email" not in user_info:
+                raise SuspiciousOperation("OIDC client did not return email address")
+            users = get_user_model().objects.filter(email__iexact=user_info["email"])
+            if len(users) == 1:
+                return self.update_user(users.first(), user_info)
+            elif len(users) > 1:
+                raise SuspiciousOperation("Multiple users with same email address")
+            else:
+                return self.create_user(user_info)
+        except (KeyError, EphiosOIDCClient.DoesNotExist, InvalidTokenError):
```
felixrindt, line 67: lol

ephios/templates/registration/login.html
```diff
@@ -9,16 +9,27 @@
 
 {% block content %}
     {% translate "Login" as login_message %}
+    {% oidc_clients as clients %}
```
felixrindt, line 12: can we build this as template var instead of a tag?

ephios/extra/auth.py
felixrindt: what if `self.client.user_endpoint` is not available and http 500?

ephios/core/migrations/0023_ephiosoidcclient_logout_url.py
```diff
@@ -0,0 +1,17 @@
...
+    dependencies = [
+        ("core", "0022_ephiosoidcclient"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="ephiosoidcclient",
+            name="logout_url",
+            field=models.URLField(blank=True, null=True),
+        ),
```
felixrindt, line 16: squash migrations in the end :)

ephios/templates/registration/login.html
jeriox: difficult, because that is the built-in django login view and not one of our own

ephios/extra/auth.py
jeriox: what was the lol about? ^^

ephios/templates/registration/login.html
felixrindt: I see

ephios/extra/auth.py
felixrindt: List of exceptions all with the same except block

jeriox: would you prefer something else? those are all exceptions that we can't do anything about at that moment and result in the user not being authenticated, so it made sense for me

felixrindt: No I'm fine with it. Just made me chuckle

ephios/core/urls.py
```diff
@@ -251,4 +252,7 @@
         WorkingHourCreateView.as_view(),
         name="workinghours_add",
     ),
+    path("oauth-test/<int:client>", OAuthRequestView.as_view(), name="oauth_test"),
```
jeriox, line 255: fix

ephios/core/views/auth.py
```diff
@@ -0,0 +1,164 @@
...
+    success_url = reverse_lazy("core:settings_oidc_create")
+    success_message = _("OIDC client successfully created.")
+
+    def get_initial(self):
+        initial = super().get_initial()
+        if "url" in self.request.GET:
+            try:
+                oidc_configuration = requests.get(
+                    urljoin(self.request.GET["url"], ".well-known/openid-configuration")
+                ).json()
```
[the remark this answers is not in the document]
felixrindt: @jeriox well you could parse the host from the url and use just that here...

ephios/core/migrations/0022_identityprovider.py
```diff
@@ -0,0 +1,103 @@
...
+                    models.ManyToManyField(
+                        blank=True,
+                        help_text="The groups that users logging in with this provider will be added to.",
+                        to="auth.group",
+                        verbose_name="default groups",
+                    ),
+                ),
+            ],
+        ),
+    ]
```
felixrindt, line 103: We should add an operation migrating the old env-based settings to a new identity provider automatically?!

ephios/core/views/auth.py
```diff
@@ -0,0 +1,167 @@
...
+                    provider.end_session_endpoint,
+                    {"post_logout_redirect_uri": settings.GET_SITE_URL()},
+                )
+                logout_url = req.url
+        auth.logout(self.request)
+        messages.info(self.request, _("Logged out successfully."))
+        return logout_url
+
+
+class IdentityProviderCreateView(SuccessMessageMixin, CreateView):
```
felixrindt, line 89: image
Template seems not right to me. missing menu entries

felixrindt: also I got logged out when creating the provider?

felixrindt (changes requested):
This is missing documentation. Esp. how to configure the Identity provider (redirect url etc.)

ephios/extra/auth.py
```diff
@@ -1,14 +1,36 @@
...
+        header = jwt.get_unverified_header(token)
+        key = jwks_client.get_signing_key(header["kid"]).key
+        decoded = jwt.decode(token, key, [header["alg"]], audience=self.provider.client_id)
+        return decoded
+
     def create_user(self, claims):
         user = get_user_model()(email=claims.get("email"))
         user.set_unusable_password()
         user.first_name = claims.get("given_name", "")
         user.last_name = claims.get("family_name", "")
```
felixrindt, line 30: I don't like how this creates an account without any name when using e.g. nextcloud. Have you thought about what to do about this?

ephios/core/views/auth.py
```diff
@@ -0,0 +1,167 @@
...
+                    urljoin(self.request.GET["url"], ".well-known/openid-configuration"), timeout=10
+                ).json()
+                config_keys = [
+                    "authorization_endpoint",
+                    "token_endpoint",
+                    "userinfo_endpoint",
+                    "end_session_endpoint",
+                    "jwks_uri",
+                ]
+                initial.update({k: oidc_configuration[k] for k in config_keys})
```
felixrindt, line 120: That's pretty nifty! :+1:

ephios/extra/auth.py
```diff
@@ -1,14 +1,36 @@
...
+class EphiosOIDCAB(ModelBackend):
+    def decode_jwt_token(self, token: str) -> Dict[str, Any]:
+        jwks_client = jwt.PyJWKClient(self.provider.jwks_uri)
+        header = jwt.get_unverified_header(token)
+        key = jwks_client.get_signing_key(header["kid"]).key
+        decoded = jwt.decode(token, key, [header["alg"]], audience=self.provider.client_id)
+        return decoded
+
     def create_user(self, claims):
         user = get_user_model()(email=claims.get("email"))
```
felixrindt, line 27: So at the moment ephios can only use email as an identifying attribute. What if people want to use oidc providers where users can change their email freely? This could be used to hijack another ephios account!

ephios/core/views/auth.py
jeriox: jep, noticed that yesterday when going to sleep ^^

jeriox: I cannot reproduce being logged out after creating the provider?

ephios/core/views/auth.py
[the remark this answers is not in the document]
jeriox: nope, because paths matter. Keycloak needs a base url of https://keycloak.example.com/realms/REALM_NAME/ to work

ephios/extra/auth.py
jeriox: the only viable option would be using the full name and doing some magic splitting there which is not ideal as well

ephios/extra/auth.py
jeriox: hm, I think we could just add a hint about it so that people are aware when cofiguring a provider. We could also go to the full length and add another model with some kind of OIDC association, but that opens another set of problems. And I think all usecases that we have seen so far don't require such a complicated setup

ephios/core/views/auth.py
felixrindt: > I cannot reproduce being logged out after creating the provider?

I probably logged out in another tab... sorry!

ephios/extra/auth.py
felixrindt: we could ask the user to provide name/date of birth when logging in for the first time?!

jeriox: sounds good. But I would keep that separate from this PR as the problem is already there and not changed by this PR. Also, we could use that to think about some kind of user onboarding flow in general

felixrindt: to be solved with #146

ephios/extra/auth.py
felixrindt: add security hint

docs/admin/additional_features/oidc.rst
```diff
@@ -17,27 +17,45 @@ You will also need to know the following information about your OIDC provider:
...
 The following additional configuration options are available:
 
 ============================== =================================================== ========================
 Value                          Usage                                               Default value
 ============================== =================================================== ========================
-OIDC_RP_SCOPES                 Scopes to request from the RP (for additional data) ``openid profile email``
-LOGOUT_REDIRECT_URL            redirect the user to the RP logout page             None (no redirect)
+scopes                         Scopes to request from the RP (for additional data) ``openid profile email``
+end_session_endpoint           redirect the user to the RP logout page             None (no redirect)
+default groups                 groups to add all users logging in with this IDP to None (no groups)
```
felixrindt, line 49: do you want to remove the "RP"?

docs/admin/additional_features/oidc.rst
```diff
@@ -17,27 +17,45 @@ You will also need to know the following information about your OIDC provider:
...
 ============================== =================================================== ========================
 
+If users are logged in exclusively, you can also hide the local login form with the appropriate settings under "ephios instance".
```
felixrindt, line 52:
```suggestion
If users are logged in exclusively using identity providers, you can also hide the local login form with the appropriate settings under "ephios instance".
```

docs/admin/additional_features/oidc.rst
```diff
@@ -17,27 +17,45 @@ You will also need to know the following information about your OIDC provider:
...
+
```
felixrindt, line 53: add a section about claims:

* email will be used to identify an account, so it must not be changeable by the user in the ID provider

we should also explain how we handle names when doing #146

felixrindt (approved):

PR #1068, opened 2023-09-14
Rework singup methods, add complex signup method

felixrindt:
#### TODO

* [ ] sortable for the editor
* [ ] proper minified vue.js
* [ ] proper csp for vue

#### planned issues

* [x] closes #619 (make the SignupMethod interface smaller and more coherent)
    * [ ] update plugin writing documentation
* [x] closes #869 (have an AbstractSignupMethod next to BaseSignupMethod)
* [ ] closes #243 (complex method)
    * [x] remove htmx if not used...
    * [ ] write user documentation
* [ ] closes #940 (sort participations in disposition view e.g. by section)
* [ ] closes #448 (reorder sections in shift form) 

consider

better Shift Form on mobile devices: #89
allow configuring shifts >24h: #1060
shift titles: #582
clone shift: #958

#### new issues

* as replacement for #619 create a PDF exporter issue

PR #1109, opened 2023-10-22
Complex signup Editor, refactor base signup method

### LINKED ISSUE #619: Refactor BaseSignupMethod
* [ ] move PDF export functionality to PDF exporter class
* [ ] consider refactoring Mixin based composition of signup methods (e.g. mixing Sections and Approval workflows)

### LINKED ISSUE #869: Signup Method configuration Form
coming from #868, we should seperate the interface of signup methods and its associated classes (config form, signup view, ...) from the base features (signup-until, built-in checkers, form fields, etc.)

We should then also add regression tests for #868

felixrindt:
This is a squashed version of the current state of #1068.

closes #619
closes #869


ephios/plugins/complexsignup/views.py
```diff
@@ -0,0 +1,93 @@
...
+                }
+                for q in qualifications
+            },
+            cls=CustomJSONEncoder,
+        )
+        return context
+
+    @classmethod
+    def as_view(cls, **initkwargs):
+        return csp_exempt(super().as_view(**initkwargs))
```
jeriox, line 93: why csp_exempt?

ephios/templates/base.html
```diff
@@ -52,6 +52,7 @@
         {% endcompress %}
 
         {% compress js %}
+            <script type="text/javascript" src="{% static "vuejs/vue.global.js" %}"></script>
```
jeriox, line 55: sure that we want to serve vue for every view?

ephios/plugins/complexsignup/templates/complexsignup/vue_editor.html
```diff
@@ -0,0 +1,374 @@
...
+                                    </button>
+                                </div>
+                            </div>
+                        </li>
+                        <li class="list-group-item list-group-item-secondary text-center">
+                            {% translate "Add items using the buttons in the list" %}
+                        </li>
+                    </ul>
+                    <div v-if="currentBlock.block_type === 'atomic'" class="mb-3 form-check d-inline-block">
+                        <input type="checkbox"
```
jeriox, line 185: does not get saved

jeriox (commented):
block editor only briefly reviewed

ephios/plugins/complexsignup/views.py
felixrindt: Ah yes I forgot about that. Vue didn't work with csp. I tried using nonce but couldn't get it to work. Can you help?

ephios/templates/base.html
felixrindt: it is cached and I thought about using vue in the signup configuration form, where you can't modify the included js.

felixrindt: so it was a deliberate decision

ephios/plugins/complexsignup/templates/complexsignup/vue_editor.html
felixrindt: good catch! Apparently I never tested that. Had the variable name in the frontend all wrong (`allowMore` instead of `allow_more`)

ephios/plugins/complexsignup/views.py
felixrindt: (can solve that in a later PR though)

jeriox (approved):

PR #1115, opened 2023-10-23
assign users to groups from OIDC claims

### LINKED ISSUE #1105: Get groups from OIDC
When using an OIDC identity provider, you can set default groups for all users logging in with this identity provider as of #1088. As an administrator, I also want groups to be retrieved from an OIDC claim so that e.g. planners can be given a role in the IDP and get the correct group assigned in ephios.

jeriox:
closes #1105


ephios/extra/auth.py
```diff
@@ -42,7 +44,24 @@ def update_user(self, user, claims):
...
+                or []
+            )
+            for group_name in groups_in_claims:
+                try:
+                    groups.append(Group.objects.get(name__iexact=group_name))
+                except Group.DoesNotExist:
+                    if self.provider.create_missing_groups:
+                        groups.append(Group.objects.create(name=group_name))
+            user.groups.set(groups)
+        elif self.provider.default_groups.exists():
```
lukasrad02, line 64: It might be unexpected from an administrator's perspective that the default groups will not be assigned when using the groups from the claim. I think you should either change this line from `elif` to an independent `if` (which should be fine, since you're using `add`for the default groups) or state in the documentation and field help texts that `default_groups` and `group_claim` are exclusive.

lukasrad02:
From my, not Python- or Django-experienced perspective, only considering the overall OIDC stuff, this PR looks good.

Allowing dots for nested access only for the groups claim but not the other ones might seem like a random choice (especially for people not knwoing/using Keycloak's `realm_access.roles` default claim name). For consistency reasons, you might consider to add support for nested claims to all other claim identifiers or drop it from the groups.

ephios/extra/auth.py
jeriox: I stated that in the documentation (see https://ephios--1115.org.readthedocs.build/en/1115/admin/additional_features/oidc.html from the readthedocs status check). For me it seemed more intuitive to have the IDP in total control of the groups if it supports groups. I'm not sure what the usecase for adding a group to _everyone_ from an IDP would be. @felixrindt thoughts?

lukasrad02: > I stated that in the documentation (see https://ephios--1115.org.readthedocs.build/en/1115/admin/additional_features/oidc.html from the readthedocs status check).

Are you referring to _The user will be a member of all groups that are listed in the claim, meaning they will be removed from groups that they have been manually assigned to in ephios and that are not listed in the claim_?

From my perspective, the default groups are rather _automatically_ assigned to all users than _manually_ assigned, that's why I thought the docs aren't mentioning it.

> I'm not sure what the usecase for adding a group to _everyone_ from an IDP would be.

When having multiple IDPs, it might be handy to automatically assign a group to each user, identifying the IDP this account is taken from. The IDP might not assign such a role, as from the IDP's perspective, every user would have this role, making it redundant.

docs/admin/additional_features/oidc.rst
```diff
@@ -41,13 +41,26 @@ JWKS_URI                       https://your-oidc-provider.com/certs
...
+You may be required to configure your identity provider to expose the groups in the identity token, please refer to the documentation of your identity provider for more information.
+It may also be neccessary to request an additional scope from the identity provider, e.g. ``groups``. Append this scope to the ``scopes`` setting if required.
+To enable the functionality, you need to provide the name of the claim that contains the group information.
+For example, if your identity provider exposes the groups in a claim called ``groups``, you would enter ``groups`` there.
+You can also access nested claims by using a dot-separated path, e.g. ``extra_information.groups`` if the group list is situated inside an ``extra_information`` dict in the identity token.
+The value of the claim must be a list of strings, where each string is the name of a group in ephios (case-insensitive).
+The user will be a member of all groups that are listed in the claim, meaning they will be removed from groups that they have been manually assigned to in ephios and that are not listed in the claim.
```
felixrindt, line 59: I don't know whether removing users from manually assigned groups is the best way to do it. Using groups in ephios that do not exist in the more abstract IDP.

Couldn't we still have a multi-select for groups managed by the OIDC, and when the groups claim is configured, the user is not added to all, but to the oidc-managed groups provided by the claim.

ephios/extra/auth.py
felixrindt: > I'm not sure what the usecase for adding a group to _everyone_ from an IDP would be. @felixrindt thoughts?

As stated in my other comment, I think your implementation is not flexible enough. Take the HPI IDP for example. FSR uses it but needs additional groups in ephios (e.g. FSR members). Also, an IDP might provide no group for some user, but we want to have them in at least one group, so default groups could still be useful.

jeriox: I think your example is not a good one, because it does not support exposing groups via claims. The feature is suited for a scenario which the FSR will most likely have soon: a Keycloak controlled by the FSR where they can set arbitrary groups, so especially one for FSR members. I think it would be really confusing to assign some groups in ephios and some in the IDP, as there is no reason to not just assign everything in the IDP. I think the usecase provided by @lukasrad02 is more suited. If I use two different IDPs, I could use the default groups to distinguish between them. (e.g for the current staging setup have one group for everyone signing in from the HPI). However, this still seems like some niche use case for me.

docs/admin/additional_features/oidc.rst
jeriox: so two different multiselects on the provider? I think that is really confusing for people configuring it. And that would make the create_missing_groups impossible if only predefined groups would be respected

ephios/extra/auth.py
jeriox: we decided to add default groups also if oidc claims are used.

docs/admin/additional_features/oidc.rst
jeriox: we decided against that

ephios/core/forms/users.py
```diff
@@ -480,3 +482,38 @@ def clean_url(self):
...
+        ]
+        widgets = {
+            "default_groups": Select2MultipleWidget,
+        }
+
+    def clean_client_secret(self):
+        client_secret = self.cleaned_data["client_secret"]
+        if client_secret == "":
+            return self.instance.client_secret
+        return client_secret
```
felixrindt, line 519: I didn't see this form being used in the create view. I think it would be good to use the same form both in update and create view, although this clean function is not suited for the create view.

jeriox: that was a deliberate decision :D i thought it would be easier to just leave it that way than custoimzing this form so that it also fits the create

felixrindt (approved):

PR #1133, opened 2023-11-07
Federation Refactor

felixrindt:
* [x] `get_absolute_url` ist broken wenn du nicht owner der application bist
* [x] den handelnden nutzer gleich mitspeichern wenn man nen invite code redeemed und mir gefällt auch nicht das in der clean_code method der RedeemInviteCodeForm gleich die models erzeugt werden. die business logic gehört nicht in diese funktion, höchstens in ne extra "save" funktion vergleichbar modelforms aber eher noch in ne extra funktion

### DIFF
--- a/ephios/plugins/federation/views/frontend.py
+++ b/ephios/plugins/federation/views/frontend.py
@@ -176,6 +176,7 @@ class RedeemInviteCodeView(StaffRequiredMixin, FormView):
     @transaction.atomic
     def _setup_federated_host(self, data):
         oauth_application = Application(
+            user=self.request.user,
             client_type=Application.CLIENT_CONFIDENTIAL,
             authorization_grant_type=Application.GRANT_AUTHORIZATION_CODE,
             redirect_uris=urljoin(data["host_url"], reverse("federation:oauth_callback")),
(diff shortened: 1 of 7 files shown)


ephios/plugins/federation/views/frontend.py
```diff
@@ -170,7 +173,44 @@ def get_form_kwargs(self):
...
+            name=response_data["host_name"],
+            url=data["host_url"],
+            access_token=response_data["access_token"],
+            oauth_application=oauth_application,
+        )
+
     def form_valid(self, form):
+        try:
+            self._setup_federated_host(form.code_data)
+        except (KeyError, HTTPError, ReadTimeout):
```
jeriox, line 208: i think this can also throw JSONDecodeError

ephios/plugins/federation/views/frontend.py
```diff
@@ -170,7 +173,44 @@ def get_form_kwargs(self):
...
+            messages.error(
+                self.request,
+                _("Could not connect to host instance. Please try again later."),
```
jeriox, line 211: maybe we want to split the error messages based on different excpetions? the "try again later" one won't help if it is anything that is not related to connectivity issues

tests/plugins/federation/test_federation_views.py
```diff
@@ -16,6 +16,14 @@ def test_federation_get_shared_events(django_app, volunteer, federation, federat
     assert federated_event.title in response.text
 
 
+def test_federation_oauth_detail_view(django_app, superuser, federation):
```
jeriox, line 19: if this should test the oauth app detail view in general i don't think this should go here. if it should test whether an oauth app is created a check for the model should be sufficient

felixrindt: This checks whether the detail view works with oauth apps created in federation. Its a regression test. I moved it to the other test file though.

jeriox (approved):

PR #1138, opened 2023-11-09
django 5.0 upgrade

felixrindt:


ephios/core/plugins.py
```diff
@@ -64,16 +64,23 @@ class PluginSignal(Signal):
...
-        return [
-            (receiver, receiver(signal=self, sender=sender, **named))
-            for receiver in super()._live_receivers(sender)
-        ]
+        old_live_receivers = self._live_receivers
+        self._live_receivers = super()._live_receivers
+        try:
+            return self.send(sender, **named)
+        finally:
+            self._live_receivers = old_live_receivers
```
felixrindt, line 83: This solution is not thread-safe! I do want to avoid contextvars, but I wouldn't wanna ship it this way.
We could just copy the `send` implementation from super (replacing the call to `_live_receivers` in it), but it grew rather complex with async support in django5.0...

felixrindt:
I did not find any more apparent problems with the upgrade, even with our multi-tenant deployment.

PR #1143, opened 2023-11-26
improve notification framework

### LINKED ISSUE #989: Notification Framework Improvements
## problem description

The current notification system has a few downsides:

* if a notification fails there is no system to retry them or discard them. They just sit around.
* Notification order is not guaranteed. When doing multiple actions like confirming and then rejecting a participation, I believe there is no code in place to make sure the notifications come in in the correct order. Also,
* obsolete notifications are still sent. When a participation doesn't exist anymore or a rejected notification for a participation that was confirmed seconds later, there is no check to make sure that the notifications content is still correct.
* ~~It's not real-time. users need to wait for the next periodic run to get the notifications.~~ accepted
* At the same time, there is no kind of batching or rate-limiting for the run_periodic task, which might lead to problems when 2 processes try to send notifications at the same time?

#### web

First of all, there is no web list of notifications, but I think there should be. They could even be marked as read if people open an email or click a link in it or they click on the push notification. 

#### email

* ~~emails really need a nice template, with an imprint~~ already done
* ~~mails should have an unsubscribe links or at least a link to the notification management page~~ already done
* ~~emails should come from a contentful from-email instead of noreply@ (reply-to header)~~ not possible
* ~~mass mailings don't use BCC. When informing many people about a new event, that creates dozens of individual emails and therefore smtp calls and therefore hits on our email reputation. Probably. we need to research more about that.~~ not true

#### non local participants

* From reading the code, guest participants don't have the correct URL in their (email) notifications. They aren't even given the option for push notifications, which would be so cool.

#### push

* I don't think users dig the settings to enable notifications. I think we should ask them when installing the PWA or after signing up for the first time or smth like that.
* push notifications don't (always) use symbols or good deep links to be actionable... There are probably more features we aren't using.
    * clicking the push notification could lead to a web version of the full notification to read it. That's probably better then directly redirecting to the disposition view if the users didn't read the whole notification or could not, because it was too long
* when using multiple ephios accounts, one can't discriminate the push notifications origin because the title always says ephios. It may be good brand placement, but maybe we should incorporate some kind of short form or site URL into the sender field. Actually,  my phone does show the origin URL. But same goes for the PWA itself by the way (at least on my device).

## What I like about the current setup

* I like the modularity of creating new notification types and new backends. doesn't even need the visitor pattern ;)
* I like how we create Notifications directly on the type object. That makes it very versatile and specific. None of those random python get_item-calls like in dynamic-preferences.
* I like that we can also send notifications to people without a user account (other kinds of participants being the only use case atm)
* The code is easy to maintain and easy to use. It does a lot for how understandable it is imho.

## Solution

I think we should rethink what a Notification object represents. Seems to me like it should not be specific to one type of notification backend but instead the act of notifying some user about something that happened, incl. fields for

* data about sending status for all the backends
* `read` (and `archived`) or something like what github/traewelling does.
* data for type objects to construct the various sending content values like
    * icon
    * from_user or from_email
    * head/subject and text body

That doesn't yet answer how to do the multi-recipient BCCs (or other kinds of multi-recipient notifications for that matter. Maybe a notification can have multiple recipients? 

## relates issues/PRs

* code that leads to https://github.com/ephios-dev/ephios/issues/959
    * regarding something being sent twice: there are lots of explicit calls to the email framework right now, even by django itself (password reset, ephios account security notification, AGB sending, ...). Do we want to continue with this or use our notification framework for *all* emails?
* an early try to https://github.com/ephios-dev/ephios/pull/655
* #372 
* #1077
* #1041

## Feedback

Do you have any thoughts on this? What do our users think/want/need?

jeriox:
closes #989 

todo for the scope of this PR:

- [x] created_at for notifications, sort by this for web and notification sending
- [x] fix event_detail URL for nonlocal participants
- [x] proper design for views
- [x] discuss: how to track read from mails
- [x] db lock?
- [x] migration of preferences for notification type?
- [x] tests


ephios/core/services/notifications/backends.py
```diff
@@ -63,9 +57,17 @@ def title(self):
         return NotImplementedError
 
     @classmethod
-    def can_send(cls, notification):
+    def sending_possible(cls, notification):
         return notification.user is not None
```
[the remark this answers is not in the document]
jeriox: this was already implemented in user_prefers_sending, we could move it up here. 
I still like the inactivation as it forces everyone to take action, but open for debate

ephios/core/services/notifications/backends.py
```diff
@@ -29,28 +31,28 @@ def enabled_notification_backends():
...
+                raise e
+            try:
+                mail_admins(
+                    "Notification sending failed",
+                    f"Exception: {e}\n{traceback.format_exc()}",
+                )
+            except smtplib.SMTPConnectError:
+                pass  # if the mail backend threw this, mail admin will probably throw this as well
+            logger.warning(f"Notification sending failed with {e}")
+    Notification.objects.filter(
```
jeriox, line 53: couldn't verify this yet due to my buggy psql setup locally

ephios/core/services/notifications/types.py
```diff
@@ -197,12 +201,26 @@ def get_actions(cls, notification):
 
 
 class ParticipationMixin:
+    @classmethod
+    def is_obsolete(cls, notification):
+        participation: AbstractParticipation = AbstractParticipation.objects.get(
+            id=notification.data.get("participation_id")
+        )
+        return participation.state != notification.data["participation_state"]
```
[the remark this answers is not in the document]
felixrindt: What if for the change back another notification was issued. Let's forget about the code complexities for now. We would have two notifications for essentially the same action needed.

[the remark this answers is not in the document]
jeriox: yeah, but in our discussion we thought about only having one notification per participation. We decided against that to have a proper history of the changes there. So I think it would be weird if we didn't issue two notiifcations only in that scenario

ephios/core/services/notifications/backends.py
```diff
@@ -151,7 +151,7 @@ def send(cls, notification):
             "icon": "/static/ephios/img/ephios-symbol-red.svg",
         }
         if actions := notification.get_actions():
-            payload["url"] = actions[0][1]
+            payload["url"] = f"{actions[0][1]}?fromNotification={notification.pk}"
```
felixrindt, line 154: Instead of appending this to every <a> tag manually, we could just do a pass over the generated html and fix the href tags. That also links inside paragraphs would be tracked.

felixrindt: You are also not checking whether the actions point to the correct url/ephios instance.

jeriox: just fixing the HTML wouldn't be enough, as this line that you are quoting here is not part of the HTML. If you want to avoid the magic string I think wrapping the get_actions() to return the fixed links would be the better approach. I already thought about it but discarded it because I thought its fine this way

ephios/core/services/notifications/types.py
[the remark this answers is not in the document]
felixrindt: I'm worried about the szneario where a notification that has been marked as obsolete becomse unobsolete while at the same time a new notification is created.

You would see these notifications (newest at the top):

* You have been confirmed (unread, not obsolete)
* You have been put to requested (obsolete)
* You have been confirmed (possibly unread, not obsolete as per your code I commented)

I'd expect that last one to be considered obsolete, as there have been newer changes to the participation state, but your code fails to detect these newer changes as the participation is in the original state again.

ephios/core/services/notifications/backends.py
felixrindt: I do mean the complete rendered E-Mail HTML, not only the message content HTML. Of course parsing the URL again would be redundant, but I think it is worth it to have it separated from normal message URL construction.

jeriox: I still don't really get it, lets talk about this at some point
The notification is only marked as read in the middleware if the notification belongs to the user that is currently logged in, so I see no attack vector there

ephios/core/services/notifications/types.py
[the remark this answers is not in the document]
jeriox: we decided to leave it that way for now and revisit it if users complain or it occurs often

ephios/core/migrations/0026_userprofile_disabled_notifications_and_more.py
```diff
@@ -0,0 +1,61 @@
...
+                profile.disabled_notifications.append([notification_type, backend])
+        profile.save()
+
+
+def revert_disabled_notifications(apps, schema_editor):
+    UserProfile = apps.get_model("core", "UserProfile")
+    db_alias = schema_editor.connection.alias
+    backends = [backend.slug for backend in enabled_notification_backends()]
+    for profile in UserProfile.all_objects.using(db_alias).all():
+        preferences = profile.preferences["notifications__notifications"]
```
jeriox, line 27: profile.preferences does not exist during the migration for whatever reason. we need to figure out how to access this

ephios/core/models/users.py
```diff
@@ -527,6 +527,9 @@ def body(self):
     def is_obsolete(self):
         return self.notification_type.is_obsolete(self)
 
+    def __str__(self):
+        return f"{self.subject} for {self.user or 'Guest'}"
```
felixrindt, line 531: translate

felixrindt: and remember we can't translate f-strings :sweat_smile:

jeriox: yeah didn't want to translate it because it never shows in any UI, thats why f-string ^^

ephios/extra/middleware.py
```diff
@@ -13,3 +13,22 @@ def __call__(self, request):
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        from ephios.core.models import Notification
+
+        response = self.get_response(request)
+        if "fromNotification" in request.GET:
+            try:
+                notification = Notification.objects.get(pk=request.GET["fromNotification"])
```
felixrindt, line 28: replace hardcoded key

ephios/core/templates/core/notification_detail.html
```diff
@@ -0,0 +1,11 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block content %}
+    <h1 class="page-header">{{ notification.subject }}</h1>
+    <p>{{ notification.body }}</p>
+    {% for label, url in notification.get_actions %}
+        <a class="btn btn-secondary" href="{{ url }}">{{ label }}</a>
+    {% endfor %}
+    <a class="btn btn-secondary" href="{% url 'core:notification_list' %}">{% translate "View other notifications" %}</a>
```
felixrindt, line 10: this last button should be visually seperated from the notification actions

ephios/core/templates/core/notification_list.html
```diff
@@ -0,0 +1,41 @@
...
+                                {% translate "Mark as read" %}
+                            </a>
+                        {% endif %}
+                        {% for label, url in notification.get_actions %}
+                            <a class="btn btn-secondary btn-sm ml-2" href="{{ url }}">{{ label }}</a>
+                        {% endfor %}
+                    </div>
+                </div>
+            </li>
+        {% endfor %}
```
felixrindt, line 39: This view should paginate notifications

felixrindt (approved):

PR #1171, opened 2024-01-02
quote user display name for email name-addr format

jeriox:
as per https://datatracker.ietf.org/doc/html/rfc5322#section-3.4, the display name in the name-addr format may not contain special characters like commas. This caused the email address parsing to fail on a user which has been created in the form "Lastname, Firstname". However, special characters are allowed in quoted strings.


ephios/core/services/notifications/backends.py
```diff
@@ -94,7 +94,7 @@ def can_send(cls, notification):
     @classmethod
     def _get_mailaddress(cls, notification):
         if notification.user:
-            return f"{notification.user.get_full_name()} <{notification.user.email}>"
+            return f'"{notification.user.get_full_name()}" <{notification.user.email}>'
```
felixrindt, line 97: please use https://docs.python.org/3/library/email.utils.html#email.utils.formataddr

felixrindt (changes requested):

felixrindt (approved):

PR #1173, opened 2024-01-05
Add dev setup instructions for mysql in docker

felixrindt:


docs/development/contributing.rst
```diff
@@ -84,4 +84,24 @@ a translation. To add them, do run ``makemessages`` from the ``data/public/stati
...
+
+Using other databases
+---------------------
+
+By default, ephios uses a sqlite database for development and PostgreSQL for production.
+
+MySQL
+''''''
+
+We do not recommend using MySQL, but if you need a mysql for development, you can use the following docker command:
```
jeriox, line 97: why do we explain how to run a MySQL (which we do not recommend) but do not explain how postgres works?

felixrindt: I had to test mysql and remembered to dump it into the docs. I agree we should add postgres as well.

jeriox (approved):

felixrindt:
continued in #?

PR #1184, opened 2024-01-27
Checkout docker tags for dynamic versioning

felixrindt:
thanks @frcroth

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -17,6 +17,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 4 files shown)


.github/workflows/docker.yml
```diff
@@ -17,6 +17,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
```
frcroth, line 22: Not needed, as fetch depth 0 already fetches tags

felixrindt: alright, I wanna keep it though :3

jeriox (approved):

PR #1187, opened 2024-01-27
highlight unread notifications, add translations

jeriox:


ephios/static/ephios/scss/ephios_custom.scss
```diff
@@ -266,3 +266,8 @@ a.badge {
 .notification-content ul {
   padding-bottom: 1rem;
 }
+
+.unread-notification {
+  border-left-color: $primary;
+  border-left-width: 0.5rem;
+}
```
felixrindt, line 273: Not bad. I think it should be
```suggestion

.notification {
  border-left-width: 0.5rem;
  border-left-color: transparent;
}

.notification-unread {
  border-left-color: $primary;
}
```
or something like that so everything stays neatly left-aligned...

jeriox: this does not work because the list-group-item is already using a slim border, so making it transparent looks ugly

felixrindt: My point is: make it left-aligned ^^
image

felixrindt (approved):

PR #1222, opened 2024-03-09
add api route to get participations for user

jeriox:


ephios/api/views/events.py
```diff
@@ -96,3 +96,11 @@ class EventViewSet(viewsets.ReadOnlyModelViewSet):
...
         .order_by("start_time")
     )
+
+
+class ParticipationSerializer(serializers.ModelSerializer):
+    state = serializers.CharField(source="get_state_display")
+
+    class Meta:
+        model = LocalParticipation
+        fields = ["id", "shift", "state", "comment"]
```
jeriox, line 106: @felixrindt I wanted to include `start_time` and `end_time` as well, but this fails. Do you have an idea what I need to do to make it work? as it should use our custom manager?

ephios/api/urls.py
```diff
@@ -15,11 +15,23 @@
...
+)
 from ephios.extra.permissions import staff_required
 
 router = routers.DefaultRouter()
 router.register(r"events", EventViewSet)
+router.register(
+    r"users/(?P<user>[\d]+)/participations", UserParticipationView, basename="user-participations"
+)
+router.register(
+    r"users/(?P<email>[\D]+)/participations",
```
felixrindt, line 31: this regex looks wrong, right? You can have digits in an email. This should really be a separate route or people would just need to search for the user ID using the email first.

ephios/api/urls.py
```diff
@@ -15,11 +15,23 @@
...
+    MailBasedUserParticipationView,
+    UserParticipationView,
+    UserProfileMeView,
+)
 from ephios.extra.permissions import staff_required
 
 router = routers.DefaultRouter()
 router.register(r"events", EventViewSet)
+router.register(
+    r"users/(?P<user>[\d]+)/participations", UserParticipationView, basename="user-participations"
```
felixrindt, line 28: Why doesn't this route show up in the endpoint directory at http://localhost:8000/api/?

felixrindt, line 28: I think we should create a "/users/" endpoint as well if we have the participations endpoint. Should be easy.

ephios/api/views/events.py
```diff
@@ -96,3 +96,11 @@ class EventViewSet(viewsets.ReadOnlyModelViewSet):
         .prefetch_related(Prefetch("shifts__participations"))
         .order_by("start_time")
     )
+
+
+class ParticipationSerializer(serializers.ModelSerializer):
+    state = serializers.CharField(source="get_state_display")
```
felixrindt, line 102: This should not be the translated state name. I'd say lets use the integer ID of the state and the english codeword optionally as well.

ephios/api/views/events.py
felixrindt: Did you even look at what ModelSerializer does? :D
It checks for properties on the model class, which we obvsl don't have if the value comes from a custom manager. We can however hook into the field creation process and just tell it to treat it like a property field:



```suggestion
        fields = ["id", "shift", "state", "comment", "start_time", "end_time"]

    def build_unknown_field(self, field_name, model_class):
        if field_name in {"start_time", "end_time"}:
            return self.build_property_field(field_name, model_class)
        return super().build_unknown_field(field_name, model_class)
```

ephios/api/urls.py
jeriox: I don't know where that is coming from, it shows up in http://localhost:8000/api/schema/

ephios/api/fields.py
```diff
@@ -0,0 +1,6 @@
+from rest_framework.fields import ChoiceField
+
+
+class ChoiceDisplayField(ChoiceField):
+    def to_representation(self, value):
+        return {"value": value, "readable_text": self.choices[value]}
```
felixrindt, line 6: I think the django-y term is `label` (instead of `readable_text`)

ephios/api/urls.py
felixrindt: could you research :finger_right: :finger_left:

ephios/api/urls.py
```diff
@@ -14,12 +14,23 @@
...
+)
 from ephios.extra.permissions import staff_required
 
 router = routers.DefaultRouter()
 router.register(r"events", EventViewSet)
+router.register(r"shifts", ShiftViewSet)
+router.register(r"users", UserViewSet)
+router.register(r"users/by_email", UserByMailView, basename="user-by-email")
+router.register(
+    r"users/(?P<user>[\d]+)/participations", UserParticipationView, basename="user-participations"
```
felixrindt, line 32: We should add a trailing `/` everywhere ;)

felixrindt (approved):

ephios/api/urls.py
jeriox: you sure? these are only the prefixes for the viewset and DRF appends additonal URL parts to construct the actual endpoints

ephios/api/urls.py
jeriox: because DRF displays hyperlinks to all list views on the router. As there is an url param in the prefix, DRF has no chance of rendering one working hyperlink to this view

ephios/api/urls.py
felixrindt: Ah well true nvmd

felixrindt (approved):

ephios/api/views/events.py
```diff
@@ -96,3 +105,30 @@ class EventViewSet(viewsets.ReadOnlyModelViewSet):
...
+            "duration",
+        ]
+
+    def build_unknown_field(self, field_name, model_class):
+        if field_name in {"start_time", "end_time"}:
+            return self.build_property_field(field_name, model_class)
+        return super().build_unknown_field(field_name, model_class)
+
+    def get_duration(self, obj):
+        return obj.end_time - obj.start_time
```
felixrindt, line 134: This is returned as a string right now? Wouldn't we want a number type for seconds?

jeriox: sure

ephios/api/views/events.py
```diff
@@ -131,4 +131,4 @@ def build_unknown_field(self, field_name, model_class):
         return super().build_unknown_field(field_name, model_class)
 
     def get_duration(self, obj):
-        return obj.end_time - obj.start_time
+        return float((obj.end_time - obj.start_time).total_seconds())
```
felixrindt, line 134: then again, why float if all of our inputs work on full seconds.

PR #1231, opened 2024-03-15
fix CustomEventParticipantNotification for placeholder

jeriox:


ephios/core/services/notifications/types.py
```diff
@@ -549,7 +549,9 @@ def send(cls, event: Event, content: str):
                         "event_id": event.id,
                         "content": content,
                         "event_title": event.title,
-                        "event_url": participant.reverse_event_detail(event),
+                        "event_url": (
+                            participant.reverse_event_detail(event) if participant.email else None
+                        ),
```
felixrindt, line 554: I think the correct solution would be to have *less* data (drop the `event_url` key and `event_title`) and calculate that when actually rendering the notifications. This solves the bug *and* moves work/queries out of the request flow *and* is more correct as the data gathering happens later *and* is not as ambiguous as the condition you proposed.

felixrindt (changes requested):

ephios/core/services/notifications/types.py
```diff
@@ -530,29 +530,34 @@ class CustomEventParticipantNotification(AbstractNotificationHandler):
...
-                        "event_id": event.id,
-                        "content": content,
-                        "event_title": event.title,
-                        "event_url": participant.reverse_event_detail(event),
-                    },
-                )
-            )
+            for participation in shift.participations.filter(
+                state__in={AbstractParticipation.States.CONFIRMED}
+            ):
```
felixrindt, line 540: what about
```python
AbstractParticipation.objects.filter(shift__event=event, state=AbstractParticipation.States.CONFIRMED)
```
to only have one for loop?

ephios/core/services/notifications/types.py
```diff
@@ -530,29 +530,34 @@ class CustomEventParticipantNotification(AbstractNotificationHandler):
...
+                if participation.participant not in participants:
+                    participant = participation.participant
```
felixrindt, line 542:
```suggestion
                participant = participation.participant
                if participant not in participants:
```

ephios/core/services/notifications/types.py
```diff
@@ -530,29 +530,34 @@ class CustomEventParticipantNotification(AbstractNotificationHandler):
...
+                    participants.add(participant)
+                    user = (
+                        participant.user if isinstance(participant, LocalUserParticipant) else None
+                    )
```
felixrindt, line 546: It's more duck-typing like, but I'd be fine if you don't go with it

```suggestion
                    user = getattr(participant, "user", None)
```

ephios/core/services/notifications/types.py
```diff
@@ -562,17 +567,15 @@ def send(cls, event: Event, content: str):
...
             )
         Notification.objects.bulk_create(notifications)
 
     @classmethod
     def get_subject(cls, notification):
-        return _("Information for your participation at {title}").format(
-            title=notification.data.get("event_title")
-        )
+        event = Event.objects.get(pk=notification.data.get("event_id"))
+        return _("Information for your participation at {title}").format(title=event.title)
```
felixrindt, line 578: I think many of our subject lines are too long...

```suggestion
        return _("Information regarding {title}").format(title=event.title)
```

ephios/core/services/notifications/types.py
```diff
@@ -581,14 +584,21 @@ def get_body(cls, notification):
     @classmethod
     def get_actions(cls, notification):
         event = Event.objects.get(pk=notification.data.get("event_id"))
+        participation = AbstractParticipation.objects.filter(
+            pk=notification.data.get("participation_id")
+        ).first()
```
felixrindt, line 589: Why not?
```suggestion
        participation = AbstractParticipation.objects.get(pk=notification.data.get("participation_id"))
```
and 
```python
make_absolute(participation.participant.reverse_event_detail(event))
```
down below? 
If there's no participation we don't need to send anything...

jeriox: because this notification is also sent to all responsibles for the event. They don't have a participation for the event

felixrindt: I see, can you please add a comment? 😅

jeriox: I mean the code that triggers that is only a few lines above, but sure ^^

ephios/core/services/notifications/types.py
jeriox: I think the current version is clearer on what actually happens/why we make this check

ephios/core/services/notifications/types.py
jeriox: nice one

felixrindt (approved):

ephios/core/services/notifications/types.py
felixrindt: made me trip up

PR #1249, opened 2024-04-01
fix API participation permissions, allow filtering by event type

### LINKED ISSUE #1246: Participation API Permissions Might be Wrong
**Describe the bug**
I expect that when a user requests `api/users/[id]/participations` they get a list of the participations that they can see of that user. Alternatively, I would expect the route only to be accessible to some higher level of privilege (e.g., admins) and then users receive a permission error (as opposed to an empty list).

In any case, I would expect an admin to get the full list.

In my testing, I've created a new user and given him all groups and permissions I could think of, see this image:

screenshot_2024-03-31_at_17 36 58_720

When this user requests `/api/users/3/participations`, they receive the following response

```
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "count": 0,
    "next": null,
    "previous": null,
    "results": []
}
```

When I access the same route with the `admin@localhost` I get the following (expected) response

```
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "count": 1,
    "next": null,
    "previous": null,
    "results": [
        {
            "id": 6,
            "shift": 1,
            "event_title": "Concert Medical Service",
            "state": {
                "value": 1,
                "label": "bestätigt"
            },
            "comment": "",
            "start_time": "2043-06-30T14:00:00Z",
            "end_time": "2043-06-30T23:00:00Z",
            "duration": 32400.0
        }
    ]
}
```

Similarly, the new user is not allowed to access the user route (e.g., `http://localhost:8001/api/users/3/`), where they receive the following even if the user exists (and returns as expected for `admin@localhost`).

```
HTTP 404 Not Found
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": "Nicht gefunden."
}
```

**To Reproduce**
1. Create a new user.
2. Give them admin permissions.
3. Access `/api/users/[id]/participations` for some existing user who has at least one participation.

**Expected behavior**
As explained above, ideally I would expect a user to see all participations in the API they would be able to see via the web interface. Alternatively, I would expect admins to see the participations of a user. In any case, I would expect permission errors (e.g., a `403`) if a user has insufficient permissions instead of an empty list.

**Environment**
State which device, operating system, browser and browser version you are using.

Reproduced both in Safari 17.2.1 and Firefox 123.0.1 on MacOS 14.2.1. Commit `5025baa91aae7fcbdf754319fc061d51a6093e9f`

jeriox:
closes #1246

### DIFF
--- a/ephios/api/filters.py
+++ b/ephios/api/filters.py
@@ -17,8 +17,12 @@ class ParticipationFilterSet(FilterSet):
     event_type = ModelChoiceFilter(
         field_name="shift__event__type", label="event type", queryset=EventType.objects.all()
     )
-    from_date = DateTimeFilter(field_name="shift__start_time", lookup_expr="gte", label="from")
-    to_date = DateTimeFilter(field_name="shift__end_time", lookup_expr="lte", label="to")
+    start_gte = DateTimeFilter(field_name="start_time", lookup_expr="gte", label="start time after")
+    start_lte = DateTimeFilter(
+        field_name="start_time", lookup_expr="lte", label="start time before"
+    )
+    end_gte = DateTimeFilter(field_name="end_time", lookup_expr="gte", label="end time after")
+    end_lte = DateTimeFilter(field_name="end_time", lookup_expr="lte", label="end time before")
 
     class Meta:
         model = LocalParticipation
--- a/ephios/api/filters.py
+++ b/ephios/api/filters.py
@@ -1,4 +1,3 @@
-from django.utils.translation import gettext_lazy as _
 from django_filters import DateTimeFilter, ModelChoiceFilter
 from django_filters.rest_framework import FilterSet
 from guardian.shortcuts import get_objects_for_user
@@ -14,11 +13,12 @@ class ParticipationPermissionFilter(BaseFilterBackend):
 
 
 class ParticipationFilterSet(FilterSet):
+    # we cannot use gettext_lazy as it breaks sphinxcontrib.openapi (https://github.com/sphinx-contrib/openapi/issues/153)
     event_type = ModelChoiceFilter(
-        field_name="shift__event__type", label=_("event type"), queryset=EventType.objects.all()
+        field_name="shift__event__type", label="event type", queryset=EventType.objects.all()
     )
-    from_date = DateTimeFilter(field_name="shift__start_time", lookup_expr="gte", label=_("from"))
-    to_date = DateTimeFilter(field_name="shift__end_time", lookup_expr="lte", label=_("to"))
+    from_date = DateTimeFilter(field_name="shift__start_time", lookup_expr="gte", label="from")
+    to_date = DateTimeFilter(field_name="shift__end_time", lookup_expr="lte", label="to")
 
     class Meta:
         model = LocalParticipation
--- a/ephios/api/filters.py
+++ b/ephios/api/filters.py
@@ -1,8 +1,25 @@
+from django.utils.translation import gettext_lazy as _
+from django_filters import DateTimeFilter, ModelChoiceFilter
+from django_filters.rest_framework import FilterSet
 from guardian.shortcuts import get_objects_for_user
 from rest_framework.filters import BaseFilterBackend
+
+from ephios.core.models import EventType, LocalParticipation
 
 
 class ParticipationPermissionFilter(BaseFilterBackend):
     def filter_queryset(self, request, queryset, view):
         events = get_objects_for_user(request.user, "core.view_event")
         return queryset.filter(shift__event__in=events)
+
+
+class ParticipationFilterSet(FilterSet):
+    event_type = ModelChoiceFilter(
+        field_name="shift__event__type", label=_("event type"), queryset=EventType.objects.all()
+    )
+    from_date = DateTimeFilter(field_name="shift__start_time", lookup_expr="gte", label=_("from"))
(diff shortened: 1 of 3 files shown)


felixrindt (approved):
Test this please

ephios/api/views/users.py
```diff
@@ -103,8 +104,8 @@ class UserByMailView(RetrieveModelMixin, GenericViewSet):
 class UserParticipationView(viewsets.ReadOnlyModelViewSet):
     serializer_class = ParticipationSerializer
     permission_classes = [IsAuthenticatedOrTokenHasScope]
-    filter_backends = [ObjectPermissionsFilter, DjangoFilterBackend]
-    filterset_fields = ["state"]
+    filter_backends = [ParticipationPermissionFilter, DjangoFilterBackend]
+    filterset_fields = ["state", "shift__event__type"]
```
jeriox, line 108: @felixrindt this means that users have to use `?shift__event__type=3` as the query param. Are we fine with that or should we try to work around it?

felixrindt: I think we should get used to building custom Filtersets. I think it will be important for Viewsets for user, shifts, events, participations etc as well:

https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html

ephios/api/filters.py
```diff
@@ -0,0 +1,25 @@
...
+        return queryset.filter(shift__event__in=events)
+
+
+class ParticipationFilterSet(FilterSet):
+    # we cannot use gettext_lazy as it breaks sphinxcontrib.openapi (https://github.com/sphinx-contrib/openapi/issues/153)
+    event_type = ModelChoiceFilter(
+        field_name="shift__event__type", label="event type", queryset=EventType.objects.all()
+    )
+    from_date = DateTimeFilter(field_name="shift__start_time", lookup_expr="gte", label="from")
+    to_date = DateTimeFilter(field_name="shift__end_time", lookup_expr="lte", label="to")
```
felixrindt, line 21: technically participations have their own start and end time. Also we shouldn't introduce "from" if we use "start" everywhere else

ephios/api/filters.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class ParticipationPermissionFilter(BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        events = get_objects_for_user(request.user, "core.view_event")
+        return queryset.filter(shift__event__in=events)
+
+
+class ParticipationFilterSet(FilterSet):
+    # we cannot use gettext_lazy as it breaks sphinxcontrib.openapi (https://github.com/sphinx-contrib/openapi/issues/153)
```
felixrindt, line 16: wild

jeriox: indeed

ephios/api/filters.py
```diff
@@ -17,8 +17,12 @@ class ParticipationFilterSet(FilterSet):
...
         field_name="shift__event__type", label="event type", queryset=EventType.objects.all()
     )
-    from_date = DateTimeFilter(field_name="shift__start_time", lookup_expr="gte", label="from")
-    to_date = DateTimeFilter(field_name="shift__end_time", lookup_expr="lte", label="to")
+    start_gte = DateTimeFilter(field_name="start_time", lookup_expr="gte", label="start time after")
+    start_lte = DateTimeFilter(
+        field_name="start_time", lookup_expr="lte", label="start time before"
+    )
+    end_gte = DateTimeFilter(field_name="end_time", lookup_expr="gte", label="end time after")
+    end_lte = DateTimeFilter(field_name="end_time", lookup_expr="lte", label="end time before")
```
felixrindt, line 25: Did you verify that this uses the `ParticipationManager` defined on `AbstractParticipation` providing the `start_time` and `end_time` fields? I'm asking esp. because this is defined on the `LocalParticipation` model.

jeriox: yes, filtering works as expected

BenBals:
@jeriox There is weird behaviour with trailing slashes. In my local testing, requesting _without_ works fine (as opposed to other routes), but requesting with, doesn't.

That is the response to `http://localhost:8001/api/users/3/participations/?limit=1&event_type=1&start_gte=2024-04-30T00%3A00%3A00Z&start_lte=2024-08-01T23%3A55%3A55Z` is
```
HTTP 200 OK
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "count": 0,
    "next": null,
    "previous": null,
    "results": []
}
```


And the response to `http://localhost:8001/api/users/3/participations/?limit=1&event_type=1&start_gte=2024-04-30T00%3A00%3A00Z&start_lte=2024-08-01T23%3A55%3A55Z/` is
```
HTTP 400 Bad Request
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "start_lte": [
        "Bitte ein gültiges Datum und Uhrzeit eingeben."
    ]
}
```

felixrindt:
> There is weird behaviour with trailing slashes.

Well you can't just put slashes in the query part of the URL (after the `?`).

I'd expect the view to work with and without the so called trailing slash in the location part though (immediately in front of the `?`)...

Issue #1259, opened 2024-04-04
Allow creating groups without members


lukasrad02:
**Is your feature request related to a problem? Please describe.**

Sometimes, I want to create a group without any members, for example to make this group responsible for events although the user accounts of the members have not been created yet.

Unfortunately, the user select is a required field on the group creation view, and therefore it is not allowed to keep it empty.

**Describe the solution you'd like**

Allow creating groups without members by making the user select not required.

**Describe alternatives you've considered**

As an alternative, I could add some arbitrary user to the group while creating it and remove it (via the user edit view) afterwards. However, this is cumbersome and, if forgotten, could lead to wrong permissions.

**Additional context**

Due to the restriction discussed above, it is also not possible to edit (rename, change permissions, …) of empty group without adding a user.

PR #1261, opened 2024-04-05
Make Date of Birth Not Required

BenBals:
# Use case
As an admin I would like to edit a user (e.g. to assign a group), but the user profile has no dob info. Then, the current implementation forces me to set a dob to save the other new settings. If I don't know the user's dob this is awkward. 

----
fixes https://github.com/ephios-dev/ephios/issues/1259


jeriox (approved):

felixrindt:
With the second commit, this also fixes #1259

PR #1264, opened 2024-04-15
allow workinghours to be filtered by event type

### LINKED ISSUE #1194: Filter working hours by type
As a planner, I want to filter the working hours by event type or manual working hours.

jeriox:
closes #1194


ephios/core/views/workinghour.py
```diff
@@ -42,23 +43,45 @@ def _get_target_user(self):
...
     permission_required = "core.view_userprofile"
 
-    def _get_working_hours_stats(self, start: Optional[date], end: Optional[date]):
-        participations = (
-            LocalParticipation.objects.filter(
-                state=LocalParticipation.States.CONFIRMED,
-                start_time__date__gte=start,
-                end_time__date__lte=end,
+    def _get_working_hours_stats(
+        self, start: Optional[date], end: Optional[date], type: Optional[EventType]
```
felixrindt, line 67: could you fix the type annotations here? start and end don't seem optional to me

felixrindt (approved):
I find it strange that the users with 0 hours just disappear from the list. This wasn't much of a problem before, I think, but now it can happen just by filtering by type...

jeriox:
> I find it strange that the users with 0 hours just disappear from the list. This wasn't much of a problem before, I think, but now it can happen just by filtering by type...

this is alread the case in the current version. users with 0 hours will not show up in the list, which I think is fine

PR #1294, opened 2024-07-02
group dependabot pip patch updates

felixrindt:
I think this would be a great way to reduce dependabot spam while retaining seperate commits for minor/major updates of our libraries.

[Docs for the groups option](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,10 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 10
+    groups:
+      patches:
+        update-types:
+          - "patch"
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:


.github/dependabot.yml
```diff
@@ -5,7 +5,11 @@ updates:
     schedule:
       interval: "daily"
```
jeriox, line 6: maybe we need to reduce this to weekly so that it can actually be grouped?

.github/dependabot.yml
```diff
@@ -5,7 +5,11 @@ updates:
...
     open-pull-requests-limit: 10
+    groups:
+      patches:
+        update-types:
+          - "patch"
```
jeriox, line 11: IMO we could also do this for minor?

jeriox (approved):
very cool, I only knew about the option to group everything so far which is a bit too much

.github/dependabot.yml
felixrindt: I was hoping it would update existing group PRs and continously keep a single PR up to date.

.github/dependabot.yml
felixrindt: I wanted to start with this, as I tend to check release notes of minor releases more thoroughly. Sometime we change some of our code when they add a new feature.

PR #1319, opened 2024-07-24
Fix create event button on empty event list

### LINKED ISSUE #1318: Cannot create new in list view if there are no upcoming events
**Describe the bug**
When there are no upcoming events that are visible to the current user, the button to create a new event is missing in the list view on the events page and is only shown in calendar and day modes.

**To Reproduce**
1. Launch an instance with no events (for example by following the [setup steps from the docs](https://docs.ephios.de/en/latest/development/contributing.html#development-setup), but in step 9, only create a superuser and do not load the full test data)
2. Log in using your just created admin account
3. Visit the events page. The list is empty and the button to create an event is missing.
4. Switch to calendar or day view. The missing button appears

**Expected behavior**
The button to create a new event should be visible to every user who has the permission to create events, no matter whether there are upcoming events or not.

**Screenshots**
The button is missing in the list view:
The "events" page of ephios in the list view showing a notice that there are no results. A button to create an event is missing.

But is visible in calendar view
The same page in calendar view. Now the button is visible although there are still no events listed.

**Environment**
Local Ephios development instance on f6aeda9bb6259ef7554fef296adb38a6cffbbceb (latest commit on main as of writing this issue). Reproducible in latest versions of Firefox and Chromium.

felixrindt:
Fix #1318

### DIFF
--- a/ephios/core/templates/core/event_list.html
+++ b/ephios/core/templates/core/event_list.html
@@ -80,24 +80,8 @@
         </form>
         {% if mode == "calendar" or mode == "day" %}
             {% if perms.core.add_event %}
-                <div class="dropdown">
-                    <button class="btn btn-primary btn-sm dropdown-toggle d-none d-md-inline-block" type="button"
-                            id="eventCreateButton"
-                            data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
-                        <span class="fa fa-plus"></span>
-                        {% translate "Add event" %}
-                    </button>
-                    <div class="dropdown-menu dropdown-menu-right" aria-labelledby="eventCreateButton">
-                        {% for eventtype in eventtypes %}
-                            <a class="dropdown-item" href="{% url "core:event_create" eventtype.pk %}">
-                                {% blocktranslate trimmed with title=eventtype.title %}
-                                    Add {{ title }}
-                                {% endblocktranslate %}</a>
-                        {% endfor %}
-                    </div>
-                </div>
+                {% include "core/fragments/event_list_add_event_button.html" %}
             {% endif %}
-
             <div class="d-flex justify-content-between mt-2">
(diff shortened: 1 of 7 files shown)


ephios/core/templates/core/event_list.html
```diff
@@ -128,21 +112,19 @@ <h3 class="mx-3">
...
         {% endif %}
-
-
     </div>
 
-    {% if mode == "calendar" %}
-        {{ calendar }}
+    {% if mode == "list" %}
+        {% include "core/fragments/event_list_list_mode.html" %}
     {% elif not event_list %}
```
jeriox, line 119: this would hide the calendar if there are no results, is that really what we want?

felixrindt: nope

ephios/core/templates/core/event_list.html
```diff
@@ -80,24 +80,8 @@ <h1>{% translate "Events" %}
...
-                    <div class="dropdown-menu dropdown-menu-right" aria-labelledby="eventCreateButton">
-                        {% for eventtype in eventtypes %}
-                            <a class="dropdown-item" href="{% url "core:event_create" eventtype.pk %}">
-                                {% blocktranslate trimmed with title=eventtype.title %}
-                                    Add {{ title }}
-                                {% endblocktranslate %}</a>
-                        {% endfor %}
-                    </div>
-                </div>
+                {% include "core/fragments/event_list_add_event_button.html" %}
```
jeriox, line 83: what was the reason again that we include the button here for calendar and day mode and below for the list view?

jeriox (approved):

ephios/core/templates/core/event_list.html
jeriox: kk, then lgtm

PR #1343, opened 2024-08-26
Add complex shift structure

felixrindt:


felixrindt:
This is a "bug" I think we should address if we ever refactor the matching service:

Ronny is designated for the upper block, so that block should be shown as full (as no one else would be able to get in there). 
Because I can't know about the designation when scoring the other position with Ronny (because that info is only available on the position he's designated for) the algorithm considers the other position valid for him too and thus sees it as acceptable putting him there to put another person on the upper block.

image

Fixing this would require even more spaghetti-wrangling the matching code - or a whole rewrite. I think waiting to refactor the whole thing is a smart move, as this is just a minor bug with it.

You could of course also argue this is intended behavior, as for instant signup, as designation is not visible in the overview, and disposition doesn't have to be final etc.

jeriox (approved):

Issue #209, opened 2020-12-21
Allow relative dates in shift configuration


jeriox:
I expect dates used in the shift configuration for the signup methods to be used as relative dates (e.g. 10 days before the event). This is especially handy for copying an event while preserving the correct relative signup date.

PR #1348, opened 2024-09-07
add vue recurrence picker

### LINKED ISSUE #195: Improve date picking for event copy
We are currently using django-recurrence for picking dates while copying an event. This has several issues:

- the JS code violates the CSP on every click because the library uses javascript:void(0)
- we cannot enforce an end date for recurrences
- the date picker is horrible
- there is no start date for a recurrence
- the picker for "every 14. of a month" is broken

We should either fix those issues in the library or write our own code. It would also be great to have the date preview updated after every change to the rules.

### LINKED ISSUE #958: Copy shift config
As a planner responsible for an event that covers multiple days with the same setup each day, I don't want to enter all the details multiple times (e.g. sections and required qualifications). Instead, I want to be able to copy an exisiting shift configuration and create a similar shift on another day.

jeriox:
closes #195 closes #958

- Implemented with Vue, similar to the signup block editor.
- Regarding the upcoming improvements in #958, the vue code is prepared to also pick times
- Getting the dates for the preview is no longer done via django, but directly in JS
- We decided against implementing a vue build step (see #1351)

ToDo
- [x] move widget to separate template
- [x] implement shift copy


jeriox:
First draft implementation:
image

- This is achieved with Vue, similar to the signup block editor. The RRULE string export is not completed yet, same thing with the submission. Once that is done it could serve as a 1:1 replacement for the old stuff.
- Regarding the upcoming improvements in #958, it would be good to have this more flexible (esp. allow datetime picking/hourly recurrences). 
- Getting the dates from the string directly in the UI (to display them to the user) would require the rrule npm package. This would also allow for the RRULE string export, meaning we wouldn't have to re-implement that
- The last two points probably mean that we need to introduce some kind of JS dependency management/build the vue components to be reused. This needs further discussion

ephios/static/ephios/js/event_copy.js
```diff
@@ -1,27 +1,145 @@
...
+                delete rule.bymonth
+                delete rule.bysetpos
+                delete rule.month_mode
+            }
+
+            function submitForm(event) {
+                event.target.submit();
+            }
+
+            function isRuleValid(rule) {
```
jeriox, line 66: this is stricter than the RFC, but I thought it would be suitable for us (e.g. end of recurrence mandatory, days of weeks/month need to be specified instead of being derived from DTSTART)

lukasrad02:
The layout is not mobile-optimized. And it might make sense to not use two columns on mobile.

Screenshot of copy event view on mobile. Many controls overflow their containers

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,231 @@
...
+                                @click="changeFreq(rule, frequencies.DAILY)">{% translate "Daily" %}</button>
+                        <button type="button" class="btn btn-outline-secondary"
+                                :class="{'active': rule.freq===2}"
+                                @click="changeFreq(rule, frequencies.WEEKLY)">{% translate "Weekly" %}</button>
+                        <button type="button" class="btn btn-outline-secondary"
+                                :class="{'active': rule.freq===1}"
+                                @click="changeFreq(rule, frequencies.MONTHLY)">{% translate "Monthly" %}</button>
+                        <button type="button" class="btn btn-outline-secondary"
+                                :class="{'active': rule.freq===0}"
+                                @click="changeFreq(rule, frequencies.YEARLY)">{% translate "Yearly" %}</button>
```
lukasrad02, line 31: `btn-outline-secondary` is a very light color. It lacks contrast and thus fails [WCAG guidelines](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast). In my opinion, it rather looks disabled than inactive.

The default bootstrap colors have much more contrast, but I don't know which other of your views might look weird if you make `secondary` darker.

Ephios:
Screenshot of the frequency buttons

Bootstrap default colors:
image

lukasrad02:
I don't know what went wrong here, but the calculated dates are incorrect (should be 2024-09-**16** and 2024-09-**23**).

image

However, when clicking copy, the events will be copied to the right dates.

ephios/extra/templates/extra/widgets/recurrence_picker.html
jeriox: well, basically every other view as the whole design is based on the bootstrap colors, so changing that is not really feasible.
Using primary instead seems to be too bold IMO

jeriox:
> I don't know what went wrong here, but the calculated dates are incorrect (should be 2024-09-**16** and 2024-09-**23**).
> 
> image
> 
> However, when clicking copy, the events will be copied to the right dates.

what timezone is your browser in? I fought a lot with that, but on my machine it was correct in the end when I was in Europe/Berlin as well.

lukasrad02:
> what timezone is your browser in? I fought a lot with that, but on my machine it was correct in the end when I was in Europe/Berlin as well.

Europe/Berlin (according to `Intl.DateTimeFormat().resolvedOptions().timeZone`).

Same behavior across Firefox (130.0), Firefox Developer Edition (130.0b9) and Chromium (128.0.6613.137) on the same machine as well as Firefox for Android (130.0). All browsers have timezone Europe/Berlin which is also set on the respective OS.

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,231 @@
...
+                    </div>
+                    <span class="col-3">[[ frequency_strings[rule.freq] ]]</span>
+                    <span v-if="rule.freq!='DAILY'" class="col-2">{% translate "on" %}</span>
+                </div>
+
+                <div class="mb-4" v-if="rule.freq===frequencies.WEEKLY">
+                    <div class="form-check form-check-inline" v-for="(day, index) in weekdays">
+                        <input class="form-check-input" type="checkbox" v-model="rule.byweekday" :value="index"
+                               :required="!rule.byweekday">
+                        <label class="form-check-label">[[ day.slice(0,2) ]]</label>
```
lukasrad02, line 55: Please connect labels to checkboxes (`id` and `for` attributes)

lukasrad02:
I'm not familiar with the recurrence spec, but my naive expectation with a recurrence like "weekly, Mondays and Tuesdays, for 2 occurrences" would be that there are 4 copies in total (two times Monday and Tuesday), but there are only two total copies (one time Monday and Tuesday).

The behavior is consistent across the preview in the frontend and the actual copies created in the backend, so I think it is correct behavior, but it might be worth explaining, for example with an ℹ️-icon after "occurrences" that shows a tooltip explaining the interpretation of the number (overall number of copies, not number of weeks).

lukasrad02:
Maybe related to the timezones: By chance, I've noticed that the following pattern causes an error "Invalid recurrence rule: unsupported RDATE parm: TZID=EUROPE/BERLIN" reproducibly:

Pattern: Starting at 2024-09-14, weekly mondays 2 occurrences, additional dates 2024-09-18 and 2024-09-19

Tested in Firefox Developer Edition and Chromium, but the error seems to come from the backend, as it is rendered as `{{ form.errors }}`. This is the recurrence string set to the server:

```
DTSTART;TZID=Europe/Berlin:20240914T000000RRULE:FREQ=WEEKLY;INTERVAL=1;BYDAY=MO;COUNT=2RDATE;TZID=Europe/Berlin:20240918T000000,20240919T000000
```

ephios/extra/templates/extra/widgets/recurrence_picker.html
felixrindt: I think we should take another look at this with the next bootstrap update. Many components expect secondary to have more contrast. For the places where we do need less, we can add another variant between it and `light`. Then it would also be themeable (think dark theme)

jeriox:
> Maybe related to the timezones: By chance, I've noticed that the following pattern causes an error "Invalid recurrence rule: unsupported RDATE parm: TZID=EUROPE/BERLIN" reproducibly:

good catch, forgot to mention that in the PR. this property is currently not supported in the library, I opened a PR to fix this

ephios/static/ephios/js/event_copy.js
```diff
@@ -1,27 +1,145 @@
...
-                        $('#rrule_occurrences').html(data.join("<br>"));
-                        return;
-                    }
+const {createApp, ref, computed} = Vue
+
+function formatDate(date_obj, sep = "-") {
+    let month = date_obj.getMonth() + 1;
+    let day = date_obj.getDate() + 1;
+    return date_obj.getFullYear() + sep + (month < 10 ? "0" : "") + month + sep + (day < 10 ? "0" : "") + day
+}
```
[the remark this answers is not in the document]
jeriox: done, could you pls check again?

ephios/extra/templates/extra/widgets/recurrence_picker.html
jeriox: @felixrindt so keep it as is for this PR?

ephios/static/ephios/js/event_copy.js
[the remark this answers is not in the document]
lukasrad02: Now working as intended

lukasrad02 (commented):
Some additional things I've noticed:

- "Yearly" rules do not work (no preview of selected days and *Invalid recurrence rule: empty string* error when clicking Copy).
- Have you considered what behavior should be desired for signup deadlines? For copies that don't go far into the future (esp. hourly), it is nice that the signup deadline does not change, while for copies that go more far into the future, keeping the original signup deadline might be infeasibly early.
  This could either be configurable when copying (keep original deadline vs. adjust deadline relative to new shift time), or adding relative deadlines to shifts in general.

jeriox:
thx for the review!

> or adding relative deadlines to shifts in general.

https://github.com/ephios-dev/ephios/issues/209

ephios/core/templates/core/shift_copy.html
```diff
@@ -0,0 +1,24 @@
...
+    {% translate "Copy shift" %}
+{% endblock %}
+
+{% block css %}
+{% endblock %}
+
+{% block content %}
+    <div class="page-header">
+        <h1>
+            {% translate "Copy shift" %} "{{ shift.event.title }}" <small>({{ shift.get_datetime_display }})</small>
```
felixrindt, line 16: I think if a user just wants *duplicate* a shift the current setup does not explain well what is going on. They would first need to figure out how to get a *single* copy then after confirming open the shift edit view again.
Therefore maybe having the default for a single copy at the original time would be good, or a seperate view for duplicating a shift out of the shift edit view that just sends you straight to the edit view of the duplicate.

ephios/core/views/event.py
```diff
@@ -609,30 +602,43 @@ def form_valid(self, form):
...
-                            forms.DateField().to_python(self.request.POST["dtstart"]),
-                            datetime.min.time(),
-                        ),
-                    ),
-                    default=lambda obj: date_format(obj, format="SHORT_DATE_FORMAT"),
-                )
-            )
-        except (TypeError, KeyError, ValidationError):
-            return HttpResponse()
+class ShiftCopyView(CustomPermissionRequiredMixin, SingleObjectMixin, FormView):
```
felixrindt, line 610: why is this not in shift.py?

ephios/core/views/event.py
```diff
@@ -609,30 +602,43 @@ def form_valid(self, form):
...
+        shift = self.object
+        duration = shift.end_time - shift.start_time
+        meeting_offset = shift.start_time - shift.meeting_time
+        shifts_to_create = []
+        for dt in form.cleaned_data["recurrence"].xafter(timezone.now(), 1000, inc=True):
+            shift = copy(shift)
+            shift.pk = None
+            shift.meeting_time = dt - meeting_offset
+            shift.start_time = dt
+            shift.end_time = dt + duration
```
felixrindt, line 633: I get the wrong times when copying shifts (UTC instead of correct timezone?!)

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,231 @@
...
+        </ul>
+        <div class="form-group">
+            <button class="btn btn-secondary ms-1" id="add-rule" @click.stop.prevent="addRule()"><span
+                class="fa fa-plus"></span> Add rule
+            </button>
+            <button class="btn btn-secondary ms-1" id="add-date" @click.stop.prevent="addDate()"><span
+                class="fa fa-plus"></span> Add date
+            </button>
+            <button type="submit" class="btn btn-primary float-end"><span
+                class="fa fa-copy"></span> {% translate "Copy" %}</button>
```
felixrindt, line 214: The submit button should not live in the widget code, but in the templates using it. Also both are missing an abort button to go back to event detail page.

ephios/core/views/event.py
```diff
@@ -609,30 +602,43 @@ def form_valid(self, form):
         messages.success(self.request, _("Event copied successfully."))
```
felixrindt, line 602: What we are doing to copy events (and shifts?!) is insufficient, as all of the models related to that event/shift configuration aren't copied (e.g. auto qualification, federation etc.). I think we should have some more elaborate signalling to allow plugins to copy their additional contents as well.

ephios/core/views/event.py
```diff
@@ -536,23 +534,18 @@ class EventCopyView(CustomPermissionRequiredMixin, SingleObjectMixin, FormView):
...
     template_name = "core/event_copy.html"
-    form_class = EventDuplicationForm
+    form_class = EventCopyForm
 
     def setup(self, request, *args, **kwargs):
         super().setup(request, *args, **kwargs)
         self.object = self.get_object()
 
+    def get_success_url(self):
+        return reverse("core:event_copy", kwargs={"pk": self.object.pk})
```
felixrindt, line 544: When confirming the vent copy view I arrive on the event list and not at event_copy and I have no clue why :sweat_smile:

ephios/static/ephios/js/event_copy.js
```diff
@@ -1,27 +1,145 @@
...
+            const frequencies = rrule.Frequency
+            const months = [{id: 1, short: "Jan", long: "January"}, {id: 2, short: "Feb", long: "February"}, {
+                id: 3, short: "Mar", long: "March"
+            }, {id: 4, short: "Apr", long: "April"}, {id: 5, short: "May", long: "May"}, {
+                id: 6, short: "Jun", long: "June"
+            }, {id: 7, short: "Jul", long: "July"}, {id: 8, short: "Aug", long: "August"}, {
+                id: 9, short: "Sep", long: "September"
+            }, {id: 10, short: "Oct", long: "October"}, {id: 11, short: "Nov", long: "November"}, {
+                id: 12, short: "Dec", long: "December"
+            }]
```
felixrindt, line 35: should these be translated?

ephios/static/ephios/js/event_copy.js
```diff
@@ -1,27 +1,145 @@
...
+                    }
+                })
+                return set
+            })
+
+            const computed_dates = computed(() => {
+                return rrule_set ? rrule_set.value.all().map(date => {
+                    return formatDate(date) + " " + (date.getUTCHours() < 10 ? "0" : "") + date.getUTCHours() + ":" + (date.getUTCMinutes() < 10 ? "0" : "") + date.getUTCMinutes()
+                }): []
+            })
```
felixrindt, line 117: can you factor out adding the leading zeros?

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,231 @@
...
+
+                <div class="mb-4" v-if="rule.freq===frequencies.WEEKLY">
+                    <div class="form-check form-check-inline" v-for="(day, index) in weekdays">
+                        <input :id="'weekday' + index" class="form-check-input" type="checkbox" v-model="rule.byweekday" :value="index"
+                               :required="!rule.byweekday">
+                        <label :for="'weekday' + index" class="form-check-label">[[ day.slice(0,2) ]]</label>
+                    </div>
+                </div>
+
+                <div class="mb-4 p-2 card" v-if="rule.freq===frequencies.MONTHLY">
```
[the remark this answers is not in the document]
jeriox: do you have a suggestion what to put on those labels? I see your point, but once filled they are quite self-explanatory, so just appending smth seems weird

ephios/static/ephios/js/event_copy.js
jeriox: why and to where?

ephios/core/templates/core/shift_copy.html
jeriox: wontfix

ephios/extra/templates/extra/widgets/recurrence_picker.html
jeriox: ahhh but that makes it rather ugly because the copy button cannot be placed where it currently is

ephios/core/templates/core/shift_copy.html
felixrindt: I don't think this needs a separate View, but just some GET params steering the view to make a single copy. I do agree we should do this if the need actually comes up.

felixrindt: I don't think this needs a separate View, but just some GET params steering the view to make a single copy. I do agree we should do this if the need actually comes up.

ephios/core/views/event.py
felixrindt: My system time is Europe/Berlin. I have no timezone issues on other sites. 🤔
If you can't reproduce it locally, I can look into it when I'm back at that computer.

ephios/extra/templates/extra/widgets/recurrence_picker.html
felixrindt: I'm sure you can make it work 🤔
The date list could be scrollable so content below the widget doesn't get flushed to the abyss.

ephios/core/views/event.py
jeriox: hm fair point. it was easier this way as handling dates in timezones different from the browser in JS is a huge pain. as it was working for Lukas as well as for me it would be cool if you could look into it a bit on your machine to save us the major rewrite ^^

ephios/extra/templates/extra/widgets/recurrence_picker.html
jeriox: well as the button is not below the date list this is not the issue here
image

ephios/static/ephios/js/event_copy.js
felixrindt: because this is hard to read

ephios/core/views/event.py
felixrindt: So it is showing the correct times in the preview:
image
but it saves the wrong time
image

felixrindt: image
and it seems my browser is timezone aware

felixrindt: I debugged through it and I think that in ShiftCopyView you are interpreting the localized times from the form input as UTC. It shoud work if you just assign the right timezone :thinking:

ephios/extra/templates/extra/widgets/recurrence_picker.html
jeriox: but i like where the confirm button is :P

but srsly, do you have an idea for how to layout this with the confirm button below so it does not look bad?

felixrindt: I think on the very right would make sense, where we have the save button in other forms as well. It makes sense looking over the date list to confirm they are correct, then hit save.

jeriox: hm that only works well for a short list of dates, I'm not really a fan of having a vertically scrollable container there. but maybe thats fine because people probably rarely have more than 20 occurrences?

ephios/core/views/event.py
jeriox: does it work now?

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,231 @@
...
+                    </div>
+                    <span class="col-3">[[ frequency_strings[rule.freq] ]]</span>
+                    <span v-if="rule.freq < frequencies.DAILY" class="col-1">{% translate "on" %}</span>
+                </div>
+
+                <div class="mb-4" v-if="rule.freq===frequencies.WEEKLY">
+                    <div class="form-check form-check-inline" v-for="(day, index) in weekdays">
+                        <input :id="'weekday' + index" class="form-check-input" type="checkbox" v-model="rule.byweekday" :value="index"
+                               :required="!rule.byweekday">
+                        <label :for="'weekday' + index" class="form-check-label">[[ day.slice(0,2) ]]</label>
```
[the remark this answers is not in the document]
jeriox: cannot reproduce

ephios/core/views/event.py
felixrindt: We can add it later if need arises, sure.

ephios/core/views/event.py
felixrindt: Nope.

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,238 @@
...
+                    </div>
+                </div>
+            </li>
+        </ul>
+        <div class="form-group">
+            <button class="btn btn-secondary ms-1" id="add-rule" @click.stop.prevent="addRule()"><span
+                class="fa fa-plus"></span> Add rule
+            </button>
+            <button class="btn btn-secondary ms-1" id="add-date" @click.stop.prevent="addDate()"><span
+                class="fa fa-plus"></span> Add date
```
felixrindt, line 215: Translate the string and provide translations

ephios/core/templates/core/shift_copy.html
```diff
@@ -0,0 +1,24 @@
...
+        </h1>
+    </div>
+    {{ form.errors }}
```
felixrindt, line 19: in both event_copy and shift_copy, these should be in an alert box like we do for other forms, right?

ephios/core/views/event.py
felixrindt: Write a test to prove it 😅

ephios/extra/templates/extra/widgets/recurrence_picker.html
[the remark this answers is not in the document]
lukasrad02: Well I do can :sweat_smile: 

Is there something I have to do to make Django clean caches of static files or similar? But thb, just from looking at the code, this cannot be working correctly, as both the checkboxes from two different "weekly" rules will have the same id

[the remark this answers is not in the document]
lukasrad02: I can reproduce the behavior on both Firefox and Chromium, both latest version.



https://github.com/user-attachments/assets/a29b2064-887f-4023-8a91-2bed9a4a55a8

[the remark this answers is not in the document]
jeriox: ahhhhh you mean the labels right? I thought your comment was about the actual checkboxes being in sync between rules, and that is not the case. I never click on labels so I didn't get what you mean the first time

ephios/static/ephios/js/event_copy.js
```diff
@@ -1,27 +1,158 @@
...
-                        $('#rrule_occurrences').html(data.join("<br>"));
-                        return;
-                    }
+const {createApp, ref, computed} = Vue
+
+function formatDate(date_obj, sep = "-") {
+    let month = date_obj.getMonth() + 1;
+    let day = date_obj.getDate();
+    return date_obj.getFullYear() + sep + (month < 10 ? "0" : "") + month + sep + (day < 10 ? "0" : "") + day
+}
```
felixrindt, line 7: imho this should be in the localized date format we use everywhere else

ephios/core/views/event.py
```diff
@@ -539,23 +536,25 @@ class EventCopyView(CustomPermissionRequiredMixin, SingleObjectMixin, FormView):
...
-            datetime.now() - timedelta(days=1),
-            datetime.now() + timedelta(days=7305),  # allow dates up to twenty years in the future
-            inc=True,
-            dtstart=datetime.combine(
-                forms.DateField().to_python(self.request.POST["start_date"]), datetime.min.time()
-            ),
-        )
-        for date in occurrences:
+        tz = timezone.get_current_timezone()
+        for date in form.cleaned_data["recurrence"].xafter(datetime.now(), 1000, inc=True):
```
felixrindt, line 557: Seems like I can preview dates in the past but events/shifts in the past wont be created! Either subtract a century from datetime.now here or disallow past dates in the frontend

ephios/extra/templates/extra/widgets/recurrence_picker.html
jeriox: I moved the buttons out, please check if you like the placement

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,219 @@
...
+                        <input type="date" class="form-control w-25 d-inline-block ms-2 me-2 form-control-sm"
+                               v-model="rule.until" @focus="rule.end_mode='UNTIL'"
+                               :required="rule.end_mode=='UNTIL'">
+                        <input v-if="pickHour" type="time" class="form-control form-control-sm d-inline-block w-25"
+                               v-model="rule.UNTIL_TIME" @focus="rule.end_mode='UNTIL'">
+
+                    </div>
+
+                    <div class="form-check">
+                        <input :id="'until' + index" class="form-check-input mt-2" type="radio" name="end_mode" v-model="rule.end_mode"
```
lukasrad02, line 165:
```suggestion
                        <input :id="'count' + index" class="form-check-input mt-2" type="radio" name="end_mode" v-model="rule.end_mode"
```

Otherwise, the labels "for" and "occurrences" do not work

ephios/core/views/event.py
lukasrad02: Now I'm facing the opposite problem: When using single dates, the times in the preview are wrong. The actual copy operation creates the copies at the correct time.

Screenshot 2024-12-13 at 12-14-12 ephios - Copy shift

What surprises me is that there is an offset of +2, while my browser timezone is set to UTC+1

lukasrad02: For reference: This does also happen when using hourly/daily/weekly/monthly/yearly rules.

> What surprises me is that there is an offset of +2, while my browser timezone is set to UTC+1

My bad. I was using an event from your test data on June 30, so an offset of +2 makes sense (daylight saving time).

felixrindt (commented):
* [ ] translation and check JS translation if applicable

ephios/core/templates/core/shift_copy.html
```diff
@@ -0,0 +1,37 @@
...
+        <h1>
+            {% translate "Copy shift" %} "{{ shift.event.title }}" <small>({{ shift.get_datetime_display }})</small>
+        </h1>
+    </div>
+    <form method="post">
+        {% csrf_token %}
+        {{ form.recurrence }}
+        <div class="float-end">
+            <a class="btn btn-secondary me-1" href="{% url "core:event_detail" shift.event.pk shift.event.get_canonical_slug %}"><span
+                class="fa fa-times"></span> {% translate "Back" %}</a>
```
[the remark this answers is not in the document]
jeriox: had that first, looked really odd with the picker buttons above

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,219 @@
+{% load i18n %}
+{% load static %}
+
+<input type="hidden" autocomplete="off" id="pick_hour" value="{{ widget.pick_hour|lower }}">
+<input type="hidden" autocomplete="off" id="original_start" value="{{ widget.original_start }}">
+<div id="vue_app" class="row mt-4">
+    <div id="chooser" class="col-12 col-md-6 mb-2">
```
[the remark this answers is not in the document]
jeriox: the button that overshoots? but why is the comment here on top?

ephios/core/views/event.py
```diff
@@ -543,23 +540,24 @@ class EventCopyView(CustomPermissionRequiredMixin, SingleObjectMixin, FormView):
     permission_required = "core.add_event"
```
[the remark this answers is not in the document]
jeriox: suggestions on how to mitigate this?

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,219 @@
...
+                                @click="changeFreq(rule, frequencies.DAILY)">{% translate "Daily" %}</button>
+                        <button type="button" class="btn btn-outline-secondary"
+                                :class="{'active': rule.freq===2}"
+                                @click="changeFreq(rule, frequencies.WEEKLY)">{% translate "Weekly" %}</button>
+                        <button type="button" class="btn btn-outline-secondary"
+                                :class="{'active': rule.freq===1}"
+                                @click="changeFreq(rule, frequencies.MONTHLY)">{% translate "Monthly" %}</button>
+                        <button type="button" class="btn btn-outline-secondary"
+                                :class="{'active': rule.freq===0}"
+                                @click="changeFreq(rule, frequencies.YEARLY)">{% translate "Yearly" %}</button>
```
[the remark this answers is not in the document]
jeriox: i mean I can use inputs instead of buttons if you prefer that, but I need to reset stuff based on the new frequency anyways, so won't change anything in regards to the function calls

ephios/extra/templates/extra/widgets/recurrence_picker.html
```diff
@@ -0,0 +1,219 @@
...
+                    <div>
+                        <button type="button" class="btn btn-danger" @click="removeRule(rule)">
+                            <i class="fas fa-trash"></i><span
+                                class="visually-hidden">{% translate "Remove" %}</span>
+                        </button>
+                    </div>
+                </div>
+
+                <div class="form-inline mb-4">
+                    <span class="d-inline">{% translate "Repeat every" %}</span>
```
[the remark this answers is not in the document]
jeriox: option 1: use "alle 5 Tag(e)"
option 2: let JS handle this translation to have plural suppoer

ephios/static/ephios/js/event_copy.js
```diff
@@ -1,27 +1,153 @@
...
+                rule.byweekday = rule.freq >= rrule.Frequency.WEEKLY ? [] : 0;
+                rule.bysetpos = 1;
+                rule.bymonthday = 1;
+                rule.bymonth = 1;
+                delete rule.month_mode;
+            }
+
+            function submitForm(event) {
+                event.target.submit();
+            }
```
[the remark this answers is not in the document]
jeriox: I just remember that it didn't work for whatever reason ^^ I wrote that code too long ago

ephios/core/views/event.py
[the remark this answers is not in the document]
felixrindt: I think it's simple. We just need to check all the relevant permissions. Add event? Done. Now:

* For the first scenario, intersect the set of groups the source event is visible for with groups the acting user can make events visible for.
* For the second scenario: That's a design decision really. Imo if we hide a button than the linked view should not be available. That should go hand in hand. We could therefore just add Change-Event permissions for the source event to the copy view. If we want users to be able to copy events they aren't responsible for, we'd just need "view_event". Checking the visible_for groups as above would mitigate the rest of the problem.

ephios/extra/templates/extra/widgets/recurrence_picker.html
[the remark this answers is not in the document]
felixrindt: Because I think having the single column layout for the viewport width the screenshot was take at would solve the problems

ephios/core/views/event.py
```diff
@@ -540,26 +538,34 @@ class EventDeleteView(CustomPermissionRequiredMixin, DeleteView):
...
+
+    def has_permission(self):
+        if not super().has_permission():
+            return False
+        can_publish_for_groups = get_objects_for_user(
+            self.request.user, "publish_event_for_group", klass=Group
         )
-        for date in occurrences:
+        visible_for = get_groups_with_perms(self.object, only_with_perms_in=["view_event"])
+        return can_publish_for_groups.intersection(visible_for.all()).exists()
```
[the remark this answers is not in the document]
felixrindt: (so change the statement in line 574)

ephios/static/ephios/scss/ephios_utilities.scss
```diff
@@ -1,11 +1,13 @@
 
-.two-columns {
-  -webkit-column-gap: 20px;
-  -moz-column-gap: 20px;
-  column-gap: 20px;
-  -webkit-column-count: 2;
-  -moz-column-count: 2;
-  column-count: 2;
+@include media-breakpoint-up(lg) {
+  .two-columns {
```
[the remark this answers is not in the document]
felixrindt: giving the date preview list some minimum height should resolve the intercolumwrap for most cases

felixrindt (approved):

PR #1361, opened 2024-09-21
add file storage

### LINKED ISSUE #504: File storage
As a planner, I expect to have a view where I can upload new files and delete previously uploaded files. 
As a user, I expect to have a view where I can view a list of uploaded files and download them.

Optionally:
As a planner, I want to attach a file from the file storage to an event.
As a user, I expect the event detail page to list the linked files and allow me to download them.

jeriox:
closes #504 

- [x] add CRUD views
- [x] discuss file type restrictions (check with python-magic?)
- [x] update deployment docs
- [x] adapt our own deployment


ephios/plugins/files/views.py
```diff
@@ -0,0 +1,60 @@
...
+            return redirect(settings.GET_USERCONTENT_URL() + request.path)
+        document = get_object_or_404(Document, id=kwargs["pk"])
+        if settings.FALLBACK_MEDIA_SERVING:
+            response = FileResponse(document.file)
+        else:
+            response = HttpResponse()
+            response["X-Accel-Redirect"] = document.file.url
+        response["Content-Disposition"] = (
+            "attachment; filename=" + os.path.split(document.file.name)[1]
+        )
```
felixrindt, line 31: This would need a comment as well. Maybe we can craft this into an `AcceleratedMediaResponse` class to make it reusable?

ephios/extra/middleware.py
```diff
@@ -36,3 +39,17 @@ def __call__(self, request):
...
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        if (
+            request.get_host() == urlsplit(settings.GET_USERCONTENT_URL()).netloc
+            and request.resolver_match.url_name != "document"
+        ):
+            return redirect(urljoin(settings.GET_SITE_URL(), request.path))
+        return response
```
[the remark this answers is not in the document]
jeriox: yeah good point. We could move the list of "media views" to the settings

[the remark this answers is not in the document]
felixrindt: I still don't understand what it does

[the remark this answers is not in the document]
jeriox: if you request the usercontent domain, but on a view that does not serve usercontent (e.g. ephiosusercontent.de/accounts/login), it will redirect you to the same view but on the regular domain. This prevents the app from being served under the content domain, and therefore on the same domain as the usercontent

ephios/extra/middleware.py
```diff
@@ -36,3 +39,20 @@ def __call__(self, request):
...
+
+class EphiosMediaFileMiddleware:
+    """Ensure only media files are served from the media domain."""
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        if (
```
felixrindt, line 52:
```suggestion
        if (
        # if the usercontent URL does not contain a domain, the request host will be checked against `None` --> no redirect loop
```

ephios/extra/utils.py
```diff
@@ -34,3 +37,16 @@ def format_anything(value):
     if value is None:
         return _("None")
     return str(value)
+
+
+def accelerated_media_response(file):
+    if getattr(settings, "FALLBACK_MEDIA_SERVING", True):
```
felixrindt, line 43:
```suggestion
    if settings.FALLBACK_MEDIA_SERVING:
```

ephios/plugins/files/models.py
```diff
@@ -0,0 +1,13 @@
...
+from ephios.core.models import Event
+
+
+class Document(models.Model):
+    title = models.CharField(max_length=255)
+    file = models.FileField(upload_to="documents/", validators=[FileExtensionValidator(["pdf"])])
+    attached_to = models.ManyToManyField(Event, related_name="documents")
+
+    def __str__(self):
+        return str(self.title)
```
felixrindt, line 13: register for logging

ephios/plugins/files/forms.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def clean_file(self):
+        file_size = self.cleaned_data["file"].size
+        if file_size > settings.GET_USERCONTENT_QUOTA():
+            raise ValidationError(
+                _("The file is too large. There are only %(quota)s available."),
+                params={"quota": filesizeformat(settings.GET_USERCONTENT_QUOTA())},
+            )
+
+
```
felixrindt, line 26: I think we get a problem with duplicate filenames. Please check!

ephios/plugins/files/views.py
```diff
@@ -0,0 +1,45 @@
...
+class DocumentView(LoginRequiredMixin, MediaViewMixin, DetailView):
+    model = Document
+
+    def get(self, request, *args, **kwargs):
+        return accelerated_media_response(self.get_object().file)
+
+
+class DocumentListView(CustomPermissionRequiredMixin, ListView):
+    model = Document
+    permission_required = "files.add_document"
```
felixrindt, line 22: Needs a way to acquire these permissions. 
To that regard I think every plain responsible should be able to upload files for their event but not be able to delete files from other users?!

ephios/settings.py
```diff
@@ -316,6 +320,19 @@ def GET_SITE_URL():
...
 
 
+def GET_USERCONTENT_URL():
+    return MEDIA_URL
+
+
+def GET_USERCONTENT_QUOTA():
+    try:
+        disk_usage = os.statvfs(MEDIA_ROOT)
+        return disk_usage.f_bavail * disk_usage.f_frsize
```
felixrindt, line 330: this reads kind of cryptic to me. Can you add e.g. an intermediate variable `availbale_size_in_bytes` or something like that to convey information about what this number is.

ephios/settings.py
```diff
@@ -316,6 +320,19 @@ def GET_SITE_URL():
...
+    except AttributeError:
+        # not available on windows
+        return None
```
felixrindt, line 333: mega upload goes brrrrrrrrrrrrrrrr

ephios/settings.py
jeriox: that is just your code that i stole from the health check :P but yeah can make that better

ephios/settings.py
jeriox: do you have a better idea on what to do here? but only relevant in dev anyway

ephios/settings.py
felixrindt: Uhh maybe factor it out as a helper function then 😜

ephios/settings.py
felixrindt: Just add some constant like 1GB. Or even better, let's use an environment variable. Non-negative values mean available disk space minus that value and positive is the limit itself or smth like that

ephios/settings.py
felixrindt: Feel free to touch the health check then

ephios/settings.py
jeriox: now works on windows as well

ephios/plugins/files/views.py
jeriox: I added a setting to the group form before your comment. I'm not sure if we should tie file permissions to events so closely. The reason to keep the files separate in the first place was the flexibility to also use them somewhere else in ephios, e.g. on a page to have a Mitgliedsantrag ready at hand or similar

ephios/plugins/files/forms.py
jeriox: django appends a random string to the filename if a file with the same name already exists

ephios/plugins/files/views.py
felixrindt: I do think it makes sense to have some general file storage, however

* This is missing a non-event related way to access files currently, right?
* I think event-specific files would clutter a list with some general purpose files. Maybe file organization needs some more thought...

ephios/plugins/files/views.py
```diff
@@ -0,0 +1,45 @@
...
+    form_class = DocumentForm
+    success_url = reverse_lazy("files:settings_document_list")
+    success_message = _("File saved successfully.")
+
+
+class DocumentDeleteView(CustomPermissionRequiredMixin, SuccessMessageMixin, DeleteView):
+    model = Document
+    permission_required = "files.delete_document"
+    success_url = reverse_lazy("files:settings_document_list")
+    success_message = _("File deleted successfully.")
```
felixrindt, line 45: If I judge my tests correctly, the media files themselves aren't deleted when deleting an instance of `Document`

ephios/plugins/files/templates/files/document_list.html
```diff
@@ -0,0 +1,33 @@
...
+                           href="{% url "files:document" document.pk %}"><span
+                            class="fa fa-download"></span> <span class="d-none d-md-inline">{% translate "Download" %}</span></a>
+                        <a class="btn btn-secondary btn-sm text-nowrap me-1"
+                           href="{% url "files:settings_document_edit" document.pk %}"><span
+                            class="fa fa-edit"></span> <span class="d-none d-md-inline">{% translate "Edit" %}</span></a>
+                        <a class="btn btn-secondary btn-sm text-nowrap me-1"
+                           href="{% url "files:settings_document_delete" document.pk %}"><span
+                            class="fa fa-trash"></span> <span class="d-none d-md-inline">{% translate "Delete" %}</span></a>
+                    </td>
+                </tr>
```
felixrindt, line 29: We should really avoid introducing new html tables in templates. We should redesign this using bootstraps grid system and implement it for pages and qualifications as well, possibly including functionality for sorting/filtering like the datalist we once had. I'd be fine with moving this to an extra issue.

ephios/settings.py
```diff
@@ -224,6 +227,8 @@
 # https://docs.djangoproject.com/en/3.0/howto/static-files/
 
 STATIC_URL = env.str("STATIC_URL", default="/static/")
+MEDIA_URL = env.str("MEDIA_URL", default="/usercontent/")
+FALLBACK_MEDIA_SERVING = env.bool("FALLBACK_MEDIA_SERVING", default=False)
```
felixrindt, line 231: So people have a working dev setup I'd suggest (similar to how django handles static file serving)

```suggestion
FALLBACK_MEDIA_SERVING = env.bool("FALLBACK_MEDIA_SERVING", default=DEBUG)
```

ephios/plugins/files/models.py
```diff
@@ -0,0 +1,25 @@
...
+from django.db import models
+
+from ephios.core.models import Event
+from ephios.modellogging.log import ModelFieldsLogConfig, register_model_for_logging
+
+
+class Document(models.Model):
+    title = models.CharField(max_length=255)
+    file = models.FileField(upload_to="documents/", validators=[FileExtensionValidator(["pdf"])])
+    attached_to = models.ManyToManyField(Event, related_name="documents")
```
felixrindt, line 11: add labels to the fields (for logging)

ephios/plugins/files/templates/files/document_list.html
```diff
@@ -0,0 +1,33 @@
...
+        <thead>
+            <tr>
+                <th>{% translate "Title" %}</th>
+                <th>{% translate "Action" %}</th>
+            </tr>
+        </thead>
+        <tbody>
+            {% for document in document_list %}
+                <tr>
+                    <td class="break-word">{{ document.title }}</td>
```
felixrindt, line 17: I think it would be great to show uploader name and last modified date in the list, e.g. to look through and see old/outdated documents.

felixrindt (changes requested):

ephios/plugins/files/templates/files/document_list.html
jeriox: yeah this is actually just copy pasta from somewhere else, I don't even remember.
I'll move it to a separate issue to get this merged soonish

ephios/settings.py
```diff
@@ -316,6 +321,17 @@ def GET_SITE_URL():
...
 
 
+def GET_USERCONTENT_URL():
+    return MEDIA_URL
+
+
+def GET_USERCONTENT_QUOTA():
+    """Returns a tuple (used, free) of the user content quota in bytes"""
+    used = sum(p.stat().st_size for p in Path(MEDIA_ROOT).rglob("*"))
+    free = shutil.disk_usage(MEDIA_ROOT).free
```
felixrindt, line 331: I still think we should have something like
```suggestion
    quota = env.int("MEDIA_FILES_DISK_QUOTA", default=-512 * 1024 * 1024)  # leave 512 MB free by default
    disk_free = shutil.disk_usage(MEDIA_ROOT).free
    free = (disk_free + quota) if quota < 0 else min(disk_free, quota - used)
```

felixrindt (approved):

ephios/settings.py
jeriox: I'm not sure if we need the complicated setup with positive and negative values, I think that is rather hard to understand. I'd prefer just setting an absolute amount that is available (as we will do in SaaS) and then do `min(disk_free, quota - used)`

felixrindt: sure :)

PR #1364, opened 2024-09-28
delete user webpush subscription if delivery fails

### LINKED ISSUE #1198: Remove push subscription if delivery fails
When delivery of a push messages is rejected (e.g. because the user blocked push messages after subscribing), our delivery fails with `WebPushException: Push failed: 401 Unauthorized`. The webpush backend should catch this and remove the push subscription for the corresponding user.

jeriox:
closes #1198


felixrindt (approved):
Seems more excessive than necessary. But I guess we don't have the info which device failed?!

jeriox:
> Seems more excessive than necessary. But I guess we don't have the info which device failed?!

exactly. the library catches expired subscriptions and only deletes them for a single device, but that info does not get propagated

PR #1374, opened 2024-10-08
Improve login view

### LINKED ISSUE #1355: Redesign login page on desktop with identity provider(s) activated
When an identity provider is activated the login page looks somewhat messy on desktop (mobile is just fine).

- order everything in column?
- make the fields narrower?
- add logo option for the identity provider?

image
image

felixrindt:
* improve visual design to fix #1355 
* add option to redirect user to SSO if a sole SSO provider is configured

in the near future I'd amend to this with the option to provide a custom logo


felixrindt:
image

jeriox:
> image

IMO this has the same problem as the current layout regarding the non-alignment of SSO button and form, just with the colums inverted. I would prefer smth as suggested in https://github.com/ephios-dev/ephios/issues/1355#issuecomment-2344253676

jeriox:
While we are at it with the login: we should provide a proper error message when entering wrong login credentials. Currently the form only says "Please enter a correct email address and password. Note that both fields may be case-sensitive."

felixrindt:
image

jeriox (approved):

Issue #1383, opened 2024-10-16
Complex signup follow-ups


jeriox:
- [ ] API
- [ ] tests 
- [ ] fallback method fails with "shift state render failed"
- [ ] production vue 
- [ ] better solution for CSP exempt vues

PR #1388, opened 2024-10-21
Fix teams structure fullness display

### LINKED ISSUE #1381: Restrict team choice if team reached maximum number of participants

At our Ephios instance, we have a shift with *Direct Confirmation* as signup mode and *Named Teams* as structure. The shift has two teams A and B that both have a maximum size set. Additionally, we have enabled *Participants must provide a preferred team*.

Participants are still able to select team A after it has reached its maximum size and due to Direct Confirmation, will be automatically dispatched into that team.

Participants should not be able to select fully occupied teams.

Whether participants can still select fully occupied teams could be configurable per shift, instead of globally altering the default behavior.

**Additional context**

*none*

### LINKED ISSUE #1382: Named team is not shown as fully occupied when signup mode ist set to Direct Confirmation
**Describe the bug**

A named team that has reached its maximum number of participants is still shown as "needs participants" if the signup mode is set to Direct Confirmation.

**To Reproduce**

1. Create an event
2. Create two shifts on that event. Both shifts should have the structure *Named Teams* with a team A with a maximum size of 1. The first shift should use the signup mode *Direct Confirmation*, the other one should be set to *Request and Confirm*
3. Sign up for both shifts and select team A during signup. Confirm signup for the shift that requires confirmation.
4. Check the checkmark icon on the right: The *Direct Confirmation* shift shows that participants are still needed, while the *Request and Confirm* shift shows to be full. (See screenshot below)

**Expected behavior**

The indicator icon should always show the correct state. In the example from above, it should show "fully occupied" for both teams.

**Screenshots**

The shift on the left side uses *Direct Confirmation* (notice the "Participate" button), while the one on the right side uses *Request and Confirm*. The checkmark icon of the left shift is wrongly stating that participants are missing.

Screenshot of what is described in step 4 above

The bug also affects teams in "enough participants" state:

Similar scenario to the previos screenshot, but both teams are now configured to allow 1-2 people. The left indicator wrongly states that participants are needed, while the right one displays the correct state.

**Environment**

Lastest ephios version (0.17.1) as well as latest commit from main branch (0609b0e)

**Additional context**

The indicator shows the correct state after manual disposition. However, this requires assigning the user to another team first.

felixrindt:
Fixes #1381. Fixes #1382. 

* [x] add discussion points to #1383 

This is just a quick fix. We're unhappy with the inconsistencies in the assignment, fullness-determination and counts between the signup structures and therefore plan to follow up with a design session (considering all possible combinations and scenarios and what behavior we expect of them), a rewrite of much of the logic touched here and user documentation.

(There is for example another bug affecting the "full"-state for atomic blocks in the complex structure, which can't be fixed without a more thoughtful rewrite of the matching algorithms.)

@lukasrad02 feel free to test this. As we discussed in our call, this should fix the issues you're currently having with your events.

### DIFF
--- a/ephios/plugins/baseshiftstructures/structure/named_teams.py
+++ b/ephios/plugins/baseshiftstructures/structure/named_teams.py
@@ -7,6 +7,7 @@ from operator import itemgetter
 from django import forms
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
+from django.utils.translation import ngettext_lazy
 
 from ephios.core.models import AbstractParticipation, Qualification
 from ephios.core.signup.disposition import BaseDispositionParticipationForm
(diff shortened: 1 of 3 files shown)


ephios/plugins/baseshiftstructures/structure/named_teams.py
```diff
@@ -109,18 +110,43 @@ def __init__(self, *args, **kwargs):
...
-                "You don't qualify for {teams}."
-            ).format(teams=", ".join(str(team["title"]) for team in unqualified))
+        help_text = ""
+        if not_qualified_teams:
+            help_text = _("You don't qualify for {teams}.").format(
+                teams=", ".join(str(team["title"]) for team in not_qualified_teams)
+            )
+        if full_teams:
+            help_text += " " + ngettext_lazy(
+                "{teams} is full.", "{teams} are full.", len(full_teams)
```
jeriox, line 146: why the len(full_teams)?

jeriox (approved):

lukasrad02:
I've noticed that there is still an inconsistency in the displayed number of missing people. Is that intended?

Screenshot from 2024-10-22 13-04-31

Scenario:

- Direct Confirmation, two named teams, each min 1, max 1, no required qualifications
- Participant "Admin" signed up for team A explicitly
- Participant "User" signed up without choosing a team
- The shift view shows 2/2 participants in total (and 0/2 for the second shift)
- The event list shows 3 missing participants, although it should only be 2

PR #1393, opened 2024-10-21
introduce DownloadTicket for files

jeriox:


ephios/plugins/files/models.py
```diff
@@ -31,6 +36,22 @@ def run_on_commit():
...
+
+
+def _token_expiration():
+    return timezone.now() + timedelta(hours=1)
+
+
+class DownloadTicket(models.Model):
+    document = models.ForeignKey(Document, on_delete=models.CASCADE)
+    expiration = models.DateTimeField(default=_token_expiration)
+    token = models.CharField(max_length=255, unique=True, default=_generate_token)
```
felixrindt, line 52: I thought these would live in the cache in the same [way django-select2 does](https://github.com/codingjoe/django-select2/blob/59798d20c65cd09f6682cc18c50d1d7e2ee4df86/django_select2/views.py#L91-L105) for its heavy widgets :thinking:

ephios/plugins/files/models.py
```diff
@@ -31,6 +36,22 @@ def run_on_commit():
...
+        random.SystemRandom().choice(string.ascii_uppercase + string.digits) for _ in range(32)
+    )
+
+
+def _token_expiration():
+    return timezone.now() + timedelta(hours=1)
+
+
+class DownloadTicket(models.Model):
+    document = models.ForeignKey(Document, on_delete=models.CASCADE)
```
felixrindt, line 50: Like this the whole setup is not transferable to other files from other models. If we introduce custom logos or profile pics etc we would need to find a different solution or add a new model each. There must be a better way.

ephios/plugins/files/models.py
jeriox: oh. that's smart.

ephios/plugins/files/views.py
```diff
@@ -1,21 +1,35 @@
...
+from ephios.plugins.files.utils import file_ticket
 
 
-class DocumentView(LoginRequiredMixin, MediaViewMixin, DetailView):
+class DownloadTicketView(MediaViewMixin, View):
+    def get(self, request, *args, **kwargs):
+        file = cache.get(self.kwargs["ticket"])
+        if file is None:
+            raise Http404()
+        return accelerated_media_response(file)
```
felixrindt, line 24: let's move this and the url to core together with the  file_ticket function

ephios/plugins/files/utils.py
```diff
@@ -0,0 +1,12 @@
...
+import string
+
+from django.core.cache import cache
+
+
+def file_ticket(file):
+    key = "".join(
+        random.SystemRandom().choice(string.ascii_uppercase + string.digits) for _ in range(32)
+    )
+    cache.set(key, file, 60 * 60)
```
felixrindt, line 11: could be more like 20 seconds?

felixrindt (approved):

PR #1394, opened 2024-10-21
add guest as url param for shared events, switch schema generation to drf-spectacular

jeriox:


ephios/api/urls.py
```diff
@@ -91,13 +90,13 @@
...
-        get_schema_view(
-            title="ephios",
-            description="ephios API",
-            version=settings.EPHIOS_VERSION,
-            url=settings.GET_SITE_URL(),
-        ),
+        SpectacularAPIView.as_view(),
         name="openapi-schema",
     ),
+    path(
```
jeriox, line 96: @felixrindt the built-in schema generation of DRF is deprecated (https://www.django-rest-framework.org/api-guide/schemas/), they recommend drf-spectacular. I rebuilt everything as is, we can decide whether we want to have a swagger view here

felixrindt: I really like the swagger view! Can it show the correct base url for the example pagination links?

ephios/plugins/federation/models.py
```diff
@@ -183,10 +183,16 @@ def all_participations(self):
...
+        return reverse(
+            "federation:shift_signup",
+            kwargs={"pk": shift.pk, "guest": self.federated_user.federated_instance.pk},
+        )
 
     def reverse_event_detail(self, event):
-        return reverse("federation:event_detail", kwargs={"pk": event.pk})
+        return reverse(
+            "federation:event_detail",
+            kwargs={"pk": event.pk, "guest": self.federated_user.federated_instance.pk},
```
felixrindt, line 194: No problem here, but did you know that using
```suggestion
            kwargs={"pk": event.pk, "guest": self.federated_user.federated_instance_id},
```
causes one db query less?

felixrindt (approved):

ephios/api/urls.py
jeriox: I couldn't find a way to customize it, I think it is intentional as this is only fixed text.
When you use the "try it out" feature, it displays the correct pagination links

PR #1401, opened 2024-10-23
Cycle serviceworker cache on user change

### LINKED ISSUE #1400: Delete page cache from serviceworker on logout.
**Describe the bug**
If I log out from ephios and access pages when offline, I can still access data from when I was logged in. 

**Expected behavior**
Cached data should only be accessible if logged in and allowed to see the content. Easiest would be to just clear the cache on signout.

felixrindt:
fixes #1400

Decided on varying the cache key depending on the user. The new service worker takes care of deleting the old cache. I'm worried about a new worker not being loaded when logging in via an identity provider though. Can you help me try @jeriox ?

### DIFF
--- a/ephios/core/templates/core/serviceworker.js
+++ b/ephios/core/templates/core/serviceworker.js
@@ -33,6 +33,17 @@ self.addEventListener('activate', event => {
     );
 });
 
+self.addEventListener("message", (event) => {
+    if (event.data === "logout") {
+        event.waitUntil(
+            caches.keys().then(cacheNames => {
+                console.log("Clearing cache after logout");
+                return Promise.all(cacheNames.map(cacheName => caches.delete(cacheName)));
+            })
+        );
+    }
+});
+
 async function markResponseAsOffline(response) {
     let content = await response.body.getReader().read().then(r => r.value);
     let body = new TextDecoder().decode(content);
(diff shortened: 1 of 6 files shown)


jeriox:
> I'm worried about a new worker not being loaded when logging in via an identity provider though. Can you help me try @jeriox ?

I'm not sure why that would be the case?

ephios/core/views/pwa.py
```diff
@@ -48,7 +48,17 @@ def get_context_data(self, **kwargs):
         context["offline_url"] = reverse("core:pwa_offline")
         # Cache name: we serve /static/ files with a cache first strategy, so
         # we need to use a new cache name when the static files change with a new ephios version.
-        context["cache_name"] = f"ephios-pwa-v{settings.EPHIOS_VERSION}"
+        # Also, we need to start a new cache for every user and permission change.
+        identity = self.request.user.id if self.request.user.is_authenticated else "anonymous"
+        permissions = set(self.request.user.get_all_permissions())
```
jeriox, line 53: could this be a performance issue? if I understand it correctly, this needs a fresh DB query for every request?

felixrindt: yes, but the serviceworker is loaded asynchronously, so it does not affect the page loads.

ephios/core/templates/core/serviceworker.js
```diff
@@ -33,6 +33,17 @@ self.addEventListener('activate', event => {
...
+self.addEventListener("message", (event) => {
+    if (event.data === "logout") {
+        event.waitUntil(
+            caches.keys().then(cacheNames => {
+                console.log("Clearing cache after logout");
+                return Promise.all(cacheNames.map(cacheName => caches.delete(cacheName)));
+            })
+        );
+    }
+});
```
felixrindt, line 45: I did test this by fixing the cache name and checking that the cache is cleared when logging out in chrome and firefox.

jeriox (approved):

PR #1402, opened 2024-10-24
add API time filters

felixrindt:
* Fixes a federation bug where event duplicates would show up in the shared list
* adds the option to filter events by start and end time e.g.
`GET /api/events/?start_time__gte=2024-04-01T07%3A00%3A00%2B02%3A00`
* use the same filters for participation

* [x] also provide those filters for the ShiftViewSet

### DIFF
--- a/ephios/api/filters.py
+++ b/ephios/api/filters.py
@@ -1,5 +1,5 @@
+import django_filters
 from django_filters import FilterSet, IsoDateTimeFilter, ModelChoiceFilter
-from django_filters.rest_framework import FilterSet
 from guardian.shortcuts import get_objects_for_user
 from rest_framework.filters import BaseFilterBackend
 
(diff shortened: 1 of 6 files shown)


ephios/api/views/events.py
```diff
@@ -80,9 +82,20 @@ class ShiftViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Shift.objects.all()
 
 
+class EventFilterSet(FilterSet):
```
jeriox, line 85: maybe this should live in filters.py?

jeriox (approved):

ephios/api/views/events.py
jeriox: sure, as long as it does not break https://github.com/fsr-de/pretix-helper-validation/blob/main/pretix_fsr_validation/signals.py#L136

do you prefer smth else than isoformat? that was just the default that django-filter showed me, so i went with it

felixrindt: Well they'd need to change the API parameter name, but we can keep a deprecated old version of the parameter if you want

jeriox: I'm also fine with removing it, but then we should highlight the breaking change in the changelog so they don't update for now as they are in the presale phase where they rely on that API

ephios/api/filters.py
```diff
@@ -12,18 +12,39 @@ def filter_queryset(self, request, queryset, view):
...
+        return queryset.filter(event__in=events)
+
+
+class StartEndTimeFilterSet(FilterSet):
+    start_time = django_filters.rest_framework.IsoDateTimeFromToRangeFilter(label="start time")
+    end_time = django_filters.rest_framework.IsoDateTimeFromToRangeFilter(label="end time")
+
+    start_gte = IsoDateTimeFilter(
+        field_name="start_time", lookup_expr="gte", label="start time greater than equals"
+    )  # deprecated 2024-10, use start_time_after
```
jeriox, line 27: maybe put the deprecation into the label so that it renders in the docs?

jeriox (approved):

PR #1403, opened 2024-10-24
move files to main nav

jeriox:


felixrindt (commented):
image

I am not ready to ship this :see_no_evil:

jeriox:
> I am not ready to ship this 🙈

well just no more german /s

image

do you have suggestions? my suggestion from earlier:
maybe einen Punkt Plugins/Module/Erweiterungen und darunter hängen dann im dropdown alle Sachen die per Signal kommen?

felixrindt:
> do you have suggestions? my suggestion from earlier: maybe einen Punkt Plugins/Module/Erweiterungen und darunter hängen dann im dropdown alle Sachen die per Signal kommen?

I think we could overflow items into the dropdown menu for the time being :thinking: 
I'm no huge fan of introducing some random category. Naming it would be hard and what goes in it kind of arbitrary.
Later maybe the whole navbar needs to be updated to some more modern concept, e.g. if we ever do some theming update.

jeriox:
> I think we could overflow items into the dropdown menu for the time being 🤔

not sure how intuitive that is, as there is only personal stuff (apart from the fact that you need to go via your personal settings to the ephios settings). I'm not sure that users will look there. Also not ideal for mobile where the main nav is fine as it stacks vertically

> I'm no huge fan of introducing some random category. Naming it would be hard and what goes in it kind of arbitrary.
well everything that would go in there would have been enabled as a plugin before, so I think extensions/plugins is reasonable. it also makes sense to me as it is everything that does not belong to the core ephios stuff

ephios/core/context.py
```diff
@@ -13,8 +15,14 @@ def ephios_base_context(request):
...
     nav = []
+    nav_groups = defaultdict(list)
     for _, result in nav_link.send(None, request=request):
-        nav += result
+        for item in result:
+            if group := item.get("group"):
+                nav_groups[group].append(item)
+            else:
+                nav.append(item)
+    nav_groups.default_factory = None  # cannot loop defaultdicts in django templates
```
felixrindt, line 25: # would that work?
```suggestion
    nav_groups =  dict(nav_groups)  # cannot loop defaultdicts in django templates
```

ephios/plugins/complexsignup/signals.py
```diff
@@ -12,3 +14,20 @@ def register_complex_shift_structures(sender, **kwargs):
...
+@receiver(nav_link, dispatch_uid="ephios.plugins.complexsignup.signals.nav_link")
+def add_nav_link(sender, request, **kwargs):
+    return (
+        [
+            {
+                "label": _("Signup blocks"),
+                "url": reverse_lazy("complexsignup:blocks_editor"),
+                "active": request.resolver_match
+                and request.resolver_match.app_name == "complexsignup",
+                "group": _("Plugins"),
```
felixrindt, line 28: I do like it, but I don't think "Plugins" is a helpful name. Also, while I do think Qualifications surpass a threshold that legitimizes its placement in the settings, but it's all very close.

ephios/plugins/complexsignup/signals.py
```diff
@@ -12,3 +14,20 @@ def register_complex_shift_structures(sender, **kwargs):
...
+            }
+        ]
+        if request.user.has_perm("core.delete_event")
+        else []
+    )
```
felixrindt, line 33: I do think we could place Working Hours, Users and Groups in their own group as well.

felixrindt (approved):

felixrindt (approved):

PR #1406, opened 2024-10-24
allow reply_to for mails

jeriox:


felixrindt (approved):
that's it?

jeriox:
> that's it?

yeah the django email class accepts reply_to as kwarg in the init method

Issue #1404, opened 2024-10-24
Improve API docs


jeriox:
- [ ] mention swagger view
- [ ] think about whether we want to keep the current rudimentary docs or host a central swagger instead
- [ ] add authorization docs to swagger
- [ ] fix drf-spectacular warnings on schema creation

PR #1412, opened 2024-11-07
API: add participation viewset

felixrindt:
@jeriox should we keep the `users/1/participations` endpoint or deprecate it and include user filtering in this one? It's not trivial, as this new ViewSet works on abstract participations and we'd need a few lines to filter for the Local user participations if a user id to filter for is given.

Also, mind that I narrowed the permission check. You now need to be able to view users **and** the event to see the participations. I'd be fine with only users too, as I find that more important in this case (includes email and stuff). Maybe we should talk about the permission model at some point. For example we could include a reduced view that only shows the VIEW_PUBLIC scoped information like in the web interface (just display name and qualifications, no email).

**Todo**

* [x] check the endpoint documentation
* [x] More tests


ephios/api/filters.py
```diff
@@ -34,15 +34,23 @@ class StartEndTimeFilterSet(FilterSet):
     )
 
 
-class ParticipationFilterSet(StartEndTimeFilterSet):
+class AbstractParticipationFilterSet(StartEndTimeFilterSet):
     # we cannot use gettext_lazy as it breaks sphinxcontrib.openapi (https://github.com/sphinx-contrib/openapi/issues/153)
```
jeriox, line 38: no more sphinxcontrib.openapi, so maybe this can be resolved

jeriox:
> @jeriox should we keep the `users/1/participations` endpoint or deprecate it and include user filtering in this one? It's not trivial, as this new ViewSet works on abstract participations and we'd need a few lines to filter for the Local user participations if a user id to filter for is given.

I think we should keep the other one. The usecase is quite different and it feels weird to include users here as it deals with AbstractParticipations

> Also, mind that I narrowed the permission check. You now need to be able to view users **and** the event to see the participations. I'd be fine with only users too, as I find that more important in this case (includes email and stuff). Maybe we should talk about the permission model at some point. For example we could include a reduced view that only shows the VIEW_PUBLIC scoped information like in the web interface (just display name and qualifications, no email).

Both sounds good.
I liked the narrowed-down version for VIEW_PUBLIC

felixrindt:
> > Also, mind that I narrowed the permission check. You now need to be able to view users **and** the event to see the participations. I'd be fine with only users too, as I find that more important in this case (includes email and stuff). Maybe we should talk about the permission model at some point. For example we could include a reduced view that only shows the VIEW_PUBLIC scoped information like in the web interface (just display name and qualifications, no email).
> 
> Both sounds good. I liked the narrowed-down version for VIEW_PUBLIC

ephios/api/access/auth.py
```diff
@@ -1,9 +1,24 @@
+from drf_spectacular.extensions import OpenApiAuthenticationExtension
+from drf_spectacular.plumbing import build_bearer_security_scheme_object
 from oauth2_provider.contrib.rest_framework import OAuth2Authentication
 
 from ephios.api.models import AccessToken
 from ephios.core.models import UserProfile
 
 
+class TokenScheme(OpenApiAuthenticationExtension):
```
jeriox, line 9: nice! you can resolve that on #1404

ephios/api/views/events.py
```diff
@@ -119,29 +67,35 @@ class EventViewSet(viewsets.ReadOnlyModelViewSet):
...
+    Remove information that would not be visible in the web version,
+    i.e. no email and date of birth of participants
+    """
 
-    def get_duration(self, obj):
-        return (obj.end_time - obj.start_time).total_seconds()
+    serializer_class = ParticipationSerializer
+    filter_backends = [ParticipationPermissionFilter, DjangoFilterBackend]
+    permission_classes = [IsAuthenticatedOrTokenHasScope]
+    required_scopes = ["CONFIDENTIAL_READ"]
```
jeriox, line 93: shouldn't that be PUBLIC_READ for the redacted version?

jeriox (approved):

ephios/api/views/events.py
felixrindt: It may be visible for logged in users, but (as per our docs) the scopes consider any people-related data as confidential.

Issue #1130, opened 2023-11-05
Improve event PDF exporter


felixrindt:
As part of the complex signup method aftermath we want to improve the pdf exporter:

- make it more userful for planners
- sorting by section
- a better programmatic interface for signup methods to customize the export

Feel free to add more suggestions to this issue

BenBals:
<img width="1003" alt="Screenshot 2024-03-09 at 14 14 31" src="https://github.com/ephios-dev/ephios/assets/2546622/afcaa00e-df1c-4b11-9576-4bb31f6256fc">

Maybe as inspiration: This is the layout we used for our last Christmas party.

jeriox:
Export as it is currently: 
We have two different layouts. One for events with only one shift:
![grafik](https://github.com/ephios-dev/ephios/assets/12702135/f9a82ea8-4131-4493-91eb-61bcd5062cd0)

And one for events with multiple shifts:
![grafik](https://github.com/ephios-dev/ephios/assets/12702135/7cc3bfc4-0881-4372-a191-d96c2f7f95cb)

While they are quite ugly, I don't think that they will become what you posted. This is mainly due to the fact that the export is only for one event at a time, while your view is more like the grid view optimised for printing. 
We already have a similar case: a client of ours has events with uniform shifts every weekend. We also built a custom PDF export for them, which layouts the whole year on a single page. This is realised by using the bulk action on the event list and implementing a custom action in a plugin. This way, you can select all the events you want to include in your export. Have a look here: https://github.com/ephios-dev/ephios-signup-beachguard/blob/main/ephios_signup_beachguard/views.py#L49

I would limit this issue to the event-specific export. @lukasrad02 and @dasGoogle also had opinions on that IIRC

dasGoogle:
Would have been smart if I had taken notes when we had this discussion... 😅 But this is what I remember: 

- Shift participants should be sorted in some predictable way (i.e. by role type, then name) as the sign-up order is not really useful for anything. 
- Regarding the Layout, the default should be A4 paper since A5 is rather untypical and does not really fit that much info
- A shift participant table should be on the same page as the shift information. I.e., rather leave some whitespace and start the whole shift on a new page
- If the table does not fit on a single page, make that obvious, since currently, it is easy to overlook a single line that overflowed to the next page
- Page number and total pages
- Some meta-information, especially when the PDF was generated would be nice. Maybe there is some more info that might be useful in a page header or so
[...]

PR #1413, opened 2024-11-08
Interface customization signals

felixrindt:
* [x] add `visible` and `force_enabled` flag to plugins and document it
* [x] add a way to customize static files/logos
* [x] make export more flexible
* [x] fix removal of sphinxcontrib.openapi
* [x] export: change email back to qualification (privacy issue?)
* [x] left align pdf content
* [x] smaller status column
* [x] add a way to add additional logos

### DIFF
--- a/ephios/api/filters.py
+++ b/ephios/api/filters.py
@@ -35,7 +35,6 @@ class StartEndTimeFilterSet(FilterSet):
 
 
 class ParticipationFilterSet(StartEndTimeFilterSet):
-    # we cannot use gettext_lazy as it breaks sphinxcontrib.openapi (https://github.com/sphinx-contrib/openapi/issues/153)
     event_type = ModelChoiceFilter(
         field_name="shift__event__type", label="event type", queryset=EventType.objects.all()
     )
--- a/ephios/api/filters.py
+++ b/ephios/api/filters.py
@@ -1,4 +1,5 @@
 import django_filters
+from django.utils.translation import gettext_lazy as _
 from django_filters import FilterSet, IsoDateTimeFilter, ModelChoiceFilter
 from guardian.shortcuts import get_objects_for_user
 from rest_framework.filters import BaseFilterBackend
@@ -19,24 +20,24 @@ class ShiftPermissionFilter(BaseFilterBackend):
 
 
 class StartEndTimeFilterSet(FilterSet):
-    start_time = django_filters.rest_framework.IsoDateTimeFromToRangeFilter(label="start time")
-    end_time = django_filters.rest_framework.IsoDateTimeFromToRangeFilter(label="end time")
+    start_time = django_filters.rest_framework.IsoDateTimeFromToRangeFilter(label=_("start time"))
+    end_time = django_filters.rest_framework.IsoDateTimeFromToRangeFilter(label=_("end time"))
 
     start_gte = IsoDateTimeFilter(
         field_name="start_time",
         lookup_expr="gte",
-        label="start time greater than equals (deprecated, use start_time_after instead)",
+        label=_("start time greater than equals (deprecated, use start_time_after instead)"),
     )
     start_lte = IsoDateTimeFilter(
         field_name="start_time",
         lookup_expr="lte",
-        label="start time less than equals (deprecated, use start_time_before instead)",
+        label=_("start time less than equals (deprecated, use start_time_before instead)"),
     )
 
 
 class ParticipationFilterSet(StartEndTimeFilterSet):
     event_type = ModelChoiceFilter(
-        field_name="shift__event__type", label="event type", queryset=EventType.objects.all()
+        field_name="shift__event__type", label=_("event type"), queryset=EventType.objects.all()
     )
 
     class Meta:
--- a/ephios/core/pdf.py
+++ b/ephios/core/pdf.py
@@ -7,12 +7,12 @@ from django.utils.translation import gettext as _
 from django.views import View
 from django.views.generic.detail import SingleObjectMixin
 from reportlab.lib import colors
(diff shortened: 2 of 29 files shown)


ephios/core/pdf.py
```diff
@@ -79,32 +148,12 @@ def get_story(self):
...
+        story.append(Paragraph(_("Participants"), self.style["Heading2"]))
+        story.append(self.get_shift_structure_data_table(shift))
         return story
 
 
-class MultipleShiftEventExporter(BasePDFExporter):
+class MultipleShiftEventExporter(EventExport):
     def __init__(self, event):
         self.event = event
         super().__init__(title=event.title)
```
jeriox, line 159: this can probably be reduced even further to have less code duplication (esp. thinking about the meeting time etc. for every shift that is defined here as well as in the single shift exporter

felixrindt: You're right, I'd defer more modifications to a seperate PR with more regards to #1130

ephios/plugins/baseshiftstructures/structure/named_teams.py
```diff
@@ -368,19 +368,47 @@ def _get_teams_with_users(self):
...
+        export_data = []
+        team_by_uuid = {team["uuid"]: team for team in self.configuration.teams}
+        for team_uuid, team in team_by_uuid.items():
+            missing_count = team["min_count"]
+            team_qualifications = (
+                Qualification.objects.filter(id=team["qualification"]) if team else []
+            )
+            for p in self.shift.participations.all():
+                participants_team_uuid = self._choose_team_for_participation(p)
+                if participants_team_uuid != team_uuid:
```
jeriox, line 381: wouldn't it be smarter to map the participations before the loop (similar to the team_by_uuid) to avoid the nested loop?

ephios/templates/base.html
```diff
@@ -72,9 +72,11 @@
         <header>
             <nav class="navbar navbar-expand-lg navbar-light bg-light shadow">
                 <div class="container">
-                    <a class="navbar-brand" href="/"><img src="{% static "ephios/img/ephios-text-black.png" %}"
-                                                          alt="ephios"
-                                                          class="d-inline-block" width="75"></a>
+                    <a class="navbar-brand" href="/">
```
jeriox, line 75: signal to customize the href would be nice for federation (can be a separate issue)

ephios/templates/base.html
```diff
@@ -116,12 +121,17 @@
...
                                             {{ user.get_full_name }}
                                         </a>
-                                        <div class="dropdown-menu" aria-labelledby="navbarDropdownAccount">
+                                        <div class="dropdown-menu dropdown-menu-end" aria-labelledby="navbarDropdownAccount">
+                                            {% for item in nav_userprofile %}
+                                                <a class="dropdown-item" {% include "core/fragments/attrs.html" with item=item %}
+                                                   href="{{ item.url }}">{{ item.label }}</a>
+                                            {% endfor %}
                                             <a class="dropdown-item"
                                                href="{% url "core:workinghours_own" %}">{% translate "Working hours" %}</a>
```
jeriox, line 136: is there a reason you left those in the HTML instead of moving it to the python code like with the sections?

jeriox (approved):

ephios/core/templates/core/fragments/attrs.html
```diff
@@ -0,0 +1 @@
+{% include "django/forms/widgets/attrs.html" with widget=item %}
```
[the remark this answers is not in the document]
felixrindt: The Django template is intended for form widgets. I first used it directly but it felt really strange assigning a nav item to a variable called widget multiple times, so I wanted to encapsulate it to have a single point of failure.

ephios/templates/base.html
felixrindt: You could specify the href using the attrs attribute 🤔

ephios/core/signals.py
```diff
@@ -50,15 +59,25 @@
...
 """
-This signal is sent out to get sections for management settings. Receivers should return a list of dicts
-containing key-value-pairs for 'label', 'url' and a boolean flag 'active'. Only views that the current user is
-allowed to view should be returned. Receivers will receive a ``request`` keyword argument.
+This signal allows you to put additional arbitrary content inside 
+the navbar. You will get the request as the keyword argument
+``request`` and are expected to return HTML.
+"""
+
+settings_sections = PluginSignal()
```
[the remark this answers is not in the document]
felixrindt: I did a search to every project I have installed locally. I think no external plugin used it (youthwarden shouldn't, beachguard is broken anyway)

ephios/templates/base.html
felixrindt: I thought it would be more work to move it. For the sections it make sense because we could automate rendering the section headings.

ephios/api/filters.py
```diff
@@ -19,25 +20,24 @@ def filter_queryset(self, request, queryset, view):
...
-        label="start time less than equals (deprecated, use start_time_before instead)",
+        label=_("start time less than equals (deprecated, use start_time_before instead)"),
     )
 
 
 class ParticipationFilterSet(StartEndTimeFilterSet):
-    # we cannot use gettext_lazy as it breaks sphinxcontrib.openapi (https://github.com/sphinx-contrib/openapi/issues/153)
     event_type = ModelChoiceFilter(
-        field_name="shift__event__type", label="event type", queryset=EventType.objects.all()
+        field_name="shift__event__type", label=_("event type"), queryset=EventType.objects.all()
```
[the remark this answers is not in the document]
felixrindt: I don't have a setup for generating the german version.

PR #1420, opened 2024-11-22
Add qualification claim to Identity Provider

### LINKED ISSUE #1352: Retrieve qualifications from OIDC IdP
We're using Ephios with Keycloak as IdP. At our Keycloak instance, we assign users to groups (representing e.g. different working groups). On login, these group memberships are applied to the Ephios account as well (see #1105), which allows us to show certain events only to a specific group of people.

However, some Ephios features use qualifications instead of group memberships, for example the *qualification mix* shift structure. Currently, our only options are to use *request and confirm* as signup flow and check eligibility manually or create multiple separate events with the respective group-based visibilities. The first option is cumbersome for us and the second one may cause confusion among managers and users.

Map qualifications from a (customizable) claim in the OIDC userinfo token. Qualifications of a user should automatically be updated each time they log in via OIDC again (just like groups do).

Combine the concepts of groups and qualifications into a single model that can be mapped from OIDC claims, used to restrict access to events (visibility, responsibility), and for shift structures.

**Additional context**
*none*

felixrindt:
Fixes #1352

design questions:

* [x] what about qualification grants that are missing in the claim?

todo

* [x] translation
* [x] tests
* [x] documentation


ephios/core/models/users.py
```diff
@@ -601,15 +607,18 @@ class IdentityProvider(Model):
         Group,
         blank=True,
         verbose_name=_("default groups"),
-        help_text=_("The groups that users logging in with this provider will be added to."),
+        help_text=_(
+            "The groups that users logging in with this provider will be added to. "
+            "Only takes effect if you don't use a group claim."
```
jeriox, line 612: FSR is using this in addition to the group claim

ephios/core/templates/core/userprofile_form.html
```diff
@@ -58,7 +62,16 @@ <h4 class="card-title">{% translate "Qualifications" %}</h4>
                                             <div class="col-md-8">
                                                 {{ form.qualification }}
                                                 {% if form.fields.qualification.disabled %}
-                                                    <span class="form-control-plaintext">{{ form.fields.qualification.title }}</span>
+                                                    <span class="form-control-plaintext">
+                                                        {{ form.fields.qualification.title }}
+                                                        {% if form.instance.externally_managed %}
+                                                            <span class="badge text-bg-danger" data-bs-toggle="tooltip"
```
jeriox, line 68: maybe text-bg-warning or -info? Danger seems a bit harsh

ephios/extra/auth.py
```diff
@@ -44,26 +49,50 @@ def update_user(self, user, claims):
...
+                target_qualification_uuids.append(
+                    uuid.UUID(
+                        str(self.provider.qualification_codename_to_uuid.get(codename, codename))
+                    )
                 )
-                or []
+            except ValueError:
+                pass
+
+        target_qualifications = Qualification.objects.filter(uuid__in=target_qualification_uuids)
```
jeriox, line 71: this makes the behaviour a bit inconsistent between qualifications (uuid-based) and groups (name-based). not sure if thats necessarily bad, just maybe not so intuitive

jeriox (approved):

ephios/core/templates/core/userprofile_form.html
felixrindt: sure :)

ephios/extra/auth.py
felixrindt: Let's see how it goes. I think I'd prefer some codename for Group names as well, so as the human readable group name can be changed freely and can use emoji etc whereas the claim can use slug-like codenames.

PR #1423, opened 2024-11-25
extend hide login button  to hide add userprofile button 

felixrindt:


ephios/core/dynamic_preferences_registry.py
```diff
@@ -54,7 +54,8 @@ class HideLoginForm(BooleanPreference):
     name = "hide_login_form"
     verbose_name = _("Hide login form")
     help_text = _(
-        "Hide the login form on the login page. This only takes effect if you configured at least one identity provider."
+        "Hide the login form on the login page and the add user button in the user list. "
```
jeriox, line 57:
```suggestion
        "Hide the login form on the login page and the controls to manage user accounts locally."
```

jeriox (approved):

ephios/core/dynamic_preferences_registry.py
felixrindt: Should we hide more controls? email field? user deletion? I don't think so. Your suggestion is still fine though.

PR #1426, opened 2024-11-26
add printing css

### LINKED ISSUE #1095: add printing css
When printing a event detail page, the layout is all over the place

felixrindt:
fix #1095

### DIFF
--- a/ephios/core/templates/core/fragments/shift_box_big.html
+++ b/ephios/core/templates/core/fragments/shift_box_big.html
@@ -19,7 +19,7 @@
         </div>
     </div>
 
-    <div class="card-footer">
+    <div class="card-footer no-print">
         {% if event.active %}
             {% with can_sign_up=request|can_sign_up:shift can_decline=request|can_decline:shift can_customize_signup=request|can_customize_signup:shift %}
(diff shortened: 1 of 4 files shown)


jeriox:
nice one!

maybe we can retain the gray background for the participant labels so that they have more contrast to the white background? and the disposition button could also use no-print

image

felixrindt:
> nice one!
> 
> maybe we can retain the gray background for the participant labels so that they have more contrast to the white background? and the disposition button could also use no-print
> 
> image
> 

That depends on the "print backgrounds" option in the print dialogue.

I do think we should look into breakpoints to avoid orphaning or windowing the shift box.

ephios/core/templates/core/fragments/shift_box_big.html
```diff
@@ -19,7 +19,7 @@
         </div>
     </div>
 
-    <div class="card-footer">
+    <div class="card-footer no-print">
```
jeriox, line 22: add for disposition button as well

jeriox (approved):

ephios/core/templates/core/fragments/shift_box_big.html
felixrindt: Thought about moving that to the card-footer.

PR #1435, opened 2024-11-28
eventtype badge visual fixes

### LINKED ISSUE #1220: Event Title is not Clickable
I was confused since the title of an event at `/events/` is not clickable. I clicked it multiple times until I found out I had to use the "Show" button.

Screenshot 2024-03-08 at 14 47 34
*Figure 1: Me selecting the event title because I tried to use it as a button.*

Clicking the name of an event has the same effect as using the "Show" button

Alternatively, move the show button to the left to make it more obvious.

**Additional context**
n/a

felixrindt:
Create  a new eventtype with a long title and you'll see visual glitches on the homepage and the event list, esp. on small screens.
This doesn't make it perfect, but it is an improvement I would say. Please try various event type name lengths and various viewport widths.

* [x] Close #1220
* [x]  check external event list

### DIFF
--- a/ephios/core/templates/core/fragments/event_list_list_mode.html
+++ b/ephios/core/templates/core/fragments/event_list_list_mode.html
@@ -32,12 +32,9 @@
                              {% if counter %}
                                  data-bs-toggle="tooltip" data-bs-placement="bottom" data-bs-html="true"
                                  title="{% for state in counter %}<div>
-
-
-
-
-
-                                            {% if event.shifts.all|length > 1 %}{{ counter|get:state }} {% endif %}{{ ParticipationStates.labels_dict|get:state }}</div>{% endfor %}"
+                                            {% if event.shifts.all|length > 1 %}
+                                                {{ counter|get:state }} {% endif %}{{ ParticipationStates.labels_dict|get:state }}
+                                            </div>{% endfor %}"
                              {% endif %}>
                             {% if counter|get:ParticipationStates.CONFIRMED > 0 %}
                                 <span class="text-success fa fa-user-check ps-2"></span>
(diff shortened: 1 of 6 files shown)


ephios/static/ephios/scss/ephios_custom.scss
```diff
@@ -1,3 +1,7 @@
+body {
+  hyphens: auto;
```
jeriox, line 2: does this break anything else? wouldn't it be enough to scope this to badges?

ephios/static/ephios/scss/ephios_custom.scss
```diff
@@ -132,9 +136,10 @@ footer {
     grid-area: time;
   }
 
-  .grid-batch {
-    grid-area: batch;
+  .grid-badge {
```
jeriox, line 139: was this a typo? :D

jeriox:
image

don't like this

ephios/static/ephios/scss/ephios_custom.scss
felixrindt: Yes

ephios/core/templates/core/fragments/event_list_list_mode.html
```diff
@@ -53,21 +58,21 @@
                             </div>
                         {% endif %}
                     </div>
-                    <a class="w-100 py-2 text-reset event-list-item-link"
-                       href="{{ event.get_absolute_url }}">
+                    <div class="w-100 py-2 event-list-item">
                         <div class="grid-wrapper m-0 py-0 ps-2 pe-3">
-                            <div class="grid-title">
+                            <a class="grid-title event-list-item-link" href="{{ event.get_absolute_url }}">
```
jeriox, line 63: is there a reason why there are multiple links instead of the event-list-item being the link?

ephios/core/templates/core/fragments/event_list_list_mode.html
```diff
@@ -93,14 +98,11 @@ <h5 class="mb-0 text-break">
                                     {% translate "to" %}
                                     {{ event.end_time|date:"SHORT_DATE_FORMAT" }}
                                 {% endif %}
-                            </div>
+                            </a>
                             <div class="grid-action d-none d-lg-flex flex-column justify-content-center">
```
jeriox, line 102: I think this div is unneccessary now? I also feel like there is too much whitespace between the signup count and the right border

jeriox (approved):

ephios/core/templates/core/fragments/event_list_list_mode.html
felixrindt: I didn't want the action are to be a link in case we want to put other links there at some point. Same as github notifications.

ephios/core/templates/core/fragments/event_list_list_mode.html
felixrindt: Yeah I do wanna keep it in case we add something there.

PR #1437, opened 2024-11-28
show buildingblock edit link only with permissions

felixrindt:
@jeriox Should we show some warning if there are no blocks?

### DIFF
--- a/ephios/core/templatetags/user_extras.py
+++ b/ephios/core/templatetags/user_extras.py
@@ -78,10 +78,9 @@ def intersects(a, b):
     return bool(set(a) & set(b))
 
 
-@register.filter(name="has_permission")
-def group_has_permission(group, permission):
-    app_label, codename = permission.split(".")
-    return group.permissions.filter(content_type__app_label=app_label, codename=codename).exists()
+@register.filter(name="user_has_permission")
+def user_has_permission(user, permission):
+    return user.has_perm(permission)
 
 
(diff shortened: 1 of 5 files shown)


ephios/core/templatetags/user_extras.py
```diff
@@ -78,10 +78,9 @@ def intersects(a, b):
     return bool(set(a) & set(b))
 
 
-@register.filter(name="has_permission")
-def group_has_permission(group, permission):
-    app_label, codename = permission.split(".")
-    return group.permissions.filter(content_type__app_label=app_label, codename=codename).exists()
```
felixrindt, line 84: this was not in use as far as I can see... Can you confirm? @jeriox

jeriox: cannot find anything either

ephios/core/templatetags/user_extras.py
```diff
@@ -78,10 +78,9 @@ def intersects(a, b):
...
+@register.filter(name="user_has_permission")
```
jeriox, line 81: is there a reason why you don't want to use the standard x in perms check in the template?

jeriox (approved):

ephios/core/templatetags/user_extras.py
felixrindt: The perms variable comes from context processors, so that variable is not available in some render contexts.

PR #1452, opened 2024-12-18
add responsible indicator

### LINKED ISSUE #874: Display event visibility
As a planner, I'd like to see for which groups an event is visible (or that it is not visible to anyone yet) on the event detail page and maybe in the event list. 

Current workarounds:
grafik
grafik

### LINKED ISSUE #1258: Also show responsible groups in the "visible for" list

When creating an event that should be visible by groups A and B and where group A is responsible/can manage the event, group A won't be shown in the "visible for" list after saving the event. This makes sense, as managing an event should always include viewing it, but it might be confusing when groups that have been added to the "visible for" list disappear from it after saving.

I'm not sure about a perfect solution, as the other way round (also showing managing groups in the "visible for" list) could also be confusing: If a group is removed from that list, but still among the managers, it will keep its view permissions, although it has been removed. Some solutions that might make sense, maybe also in combination:

- Below the "visible for" box, there is some help text saying
  
  > [...] Unabhängig davon ist die Veranstaltung für alle Benutzer sichtbar, die sich bereits angemeldet haben.
  
  It would make sense to add
  
  > [...] und alle Gruppen, die für die Veranstaltung verantwortlich sind
- When a group is added to both lists during editing, show a information sign on the group with a tooltip specifying that the group will receive the view permission automatically.
- Instead of the two separate selects, show a table of all groups with radio buttons for permissions. If a table of all groups would be to long, show an empty table with a select to add groups

*see above*

**Additional context**

*none*

felixrindt:
when testing the bulk actions, it was helpful to have some kind of indicator in the event list list to see what events I am responsible for.
This is what I came up with in a few minutes of tinkering with it.

image

closes #1258
closes #874

### DIFF
--- a/ephios/core/templates/core/fragments/event_list_list_mode.html
+++ b/ephios/core/templates/core/fragments/event_list_list_mode.html
@@ -28,25 +28,37 @@
             {% with counter=event|event_list_signup_state_counts stats=event.get_signup_stats %}
                 <li class="list-group-item p-0 d-flex flex-row">
                     <div class="m-0 py-2 d-flex flex-column flex-lg-row-reverse justify-content-around flex-grow-0">
-                        <div class="ps-lg-2 d-flex flex-row flex-lg-column justify-content-center event-list-status-icon"
-                             {% if counter %}
-                                 data-bs-toggle="tooltip" data-bs-placement="bottom" data-bs-html="true"
-                                 title="{% for state in counter %}<div>
-                                            {% if event.shifts.all|length > 1 %}
-                                                {{ counter|get:state }} {% endif %}{{ ParticipationStates.labels_dict|get:state }}
-                                            </div>{% endfor %}"
-                             {% endif %}>
-                            {% if counter|get:ParticipationStates.CONFIRMED > 0 %}
-                                <span class="text-success fa fa-user-check ps-2"></span>
-                            {% elif counter|get:ParticipationStates.REQUESTED > 0 %}
-                                <span class="text-warning fa fa-user-clock ps-2"></span>
-                            {% elif counter|get:ParticipationStates.RESPONSIBLE_REJECTED > 0 %}
-                                <span class="text-danger fa fa-user-times ps-2"></span>
-                            {% elif counter|get:ParticipationStates.USER_DECLINED > 0 and counter|length == 1 %}
-                                <span class="text-danger fa fa-user-minus ps-2"></span>
-                            {% else %}
-                                <span class="text-body-secondary far fa-user ps-2"></span>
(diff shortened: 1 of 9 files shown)


jeriox:
cool idea. first thoughts:
- I'm not sure if the icon placement is perfect, this makes the title wrap even quicker
- not sure if the crown is the best icon choice/super intuitive
- we should also think about this in combination with #874, maybe we can even come up with some kind of combined indicator?

felixrindt:
image

I got the query working :)

ephios/core/views/event.py
```diff
@@ -232,14 +233,25 @@ def get_queryset(self):
...
         else:
             # safeguard for not loading too many events
             qs = qs.filter(end_time__gte=timezone.now()).order_by("start_time", "end_time")
+        qs = qs.prefetch_related("shifts__participations")
+
+        # annotate groups that can view the event by prefetching the object permission model
+        qs = qs.prefetch_related(
+            Prefetch(
+                "group_object_permission_set",
+                queryset=GroupObjectPermission.objects.filter(permission__codename="view_event"),
```
jeriox, line 248: this includes all group with change permissons, right? probably related to #1258

felixrindt: It does, yes.

felixrindt: I'll update the help texts of the event permission fields.
I'll remove groups with global edit/view permissions from the values of these fields.

felixrindt: I'll remove groups with change permissions from the groups shown in the indicator to make it consistent with the form field.

ephios/core/templates/core/fragments/event_list_list_mode.html
```diff
@@ -28,25 +28,37 @@
...
-                            {% elif counter|get:ParticipationStates.REQUESTED > 0 %}
-                                <span class="text-warning fa fa-user-clock ps-2"></span>
-                            {% elif counter|get:ParticipationStates.RESPONSIBLE_REJECTED > 0 %}
-                                <span class="text-danger fa fa-user-times ps-2"></span>
-                            {% elif counter|get:ParticipationStates.USER_DECLINED > 0 and counter|length == 1 %}
-                                <span class="text-danger fa fa-user-minus ps-2"></span>
-                            {% else %}
-                                <span class="text-body-secondary far fa-user ps-2"></span>
+                        <div class="ps-lg-2 d-flex flex-column justify-content-center">
+                            {% if event.can_change %}
```
felixrindt, line 32: change color to info if can_change, move visible for groups to the right as a tooltip on the user-group icon.

jeriox (approved):

PR #1453, opened 2024-12-22
Option to make participant data anonymous

felixrindt:
* [x] user docs
* [x] translate
* [x] frontend
* [x] model
* [x] api
  * [x] tests


ephios/plugins/baseshiftstructures/templates/baseshiftstructures/participation_card_inline.html
```diff
@@ -2,9 +2,14 @@
...
 {% load i18n %}
 
+
 <div id="participant-popover-title-{{ participation.id }}" class="d-none">
     <span class="text-dark pe-1">{{ participation.participant.icon }}</span>
-    {{ participation.participant }}
+    {% if  participation.show_participant_data %}
+        {{ participation.participant }}
+    {% else %}
+        {% translate "Anonymous" %}
```
jeriox, line 11: I think we should use a more friendly term, IMO its weird to see that you will share the shift with "Anonymous". One idea would be to disclose the relevant qualifications (reads as: I will share the KTW with another RS). If we don't want to expose this as well we should at least use a more friendly term

ephios/plugins/baseshiftstructures/templates/baseshiftstructures/participation_card_inline.html
```diff
@@ -53,5 +60,9 @@
     {% else %}
         <i class="far fa-clock"></i>
     {% endif %}
-    {{ participation.participant }}
-</a>
\ No newline at end of file
+    {% if  participation.show_participant_data %}
+        {{ participation.participant }}
+    {% else %}
+        <i class="fas fa-user-secret px-1"></i>
```
jeriox, line 66: again not sure if that is the best visualization for your fellow volunteers
could also look a bit weird because the badges are suddenly very short?

felixrindt:
@jeriox Do you think we should show all participant info if the viewing user has view_userprofile permissions? In the API they would be allowed to view all participations of all users. I don't like the inconsistency...

jeriox:
> @jeriox Do you think we should show all participant info if the viewing user has view_userprofile permissions? In the API they would be allowed to view all participations of all users. I don't like the inconsistency...

yeah I think so. viewing the userprofile includes viewing working hours, which in turn reveals participations. so they should be visible on the event page as well.

ephios/plugins/baseshiftstructures/templates/baseshiftstructures/participation_card_inline.html
felixrindt: I agree that the wording and icon choice aren't optimal. Lets discuss this in a call?!

ephios/locale/de/LC_MESSAGES/django.po
```diff
@@ -861,140 +864,156 @@ msgstr "Dienst"
...
 
-#: ephios/core/models/events.py:49
+#: ephios/core/models/events.py:56
 msgid "color"
 msgstr "Farbe"
 
-#: ephios/core/models/events.py:53
+#: ephios/core/models/events.py:58
+msgid "show participant data"
+msgstr "Zeige Teilnehmendeninformationen"
```
felixrindt, line 890:
```suggestion
msgstr "Teilnehmendeninformationen sichtbar machen für"
```

ephios/core/models/events.py
```diff
@@ -45,8 +47,18 @@ def get_queryset(self):
...
 class EventType(Model):
+    class ShowParticipantDataChoices(models.IntegerChoices):
+        EVERYONE = 0, _("to everyone")
+        CONFIRMED = 1, _("to confirmed participants")
+        RESPONSIBLES = 2, _("only to responsible users")
+
     title = CharField(_("title"), max_length=254)
     color = CharField(_("color"), max_length=7, default="#343a40")
+    show_participant_data = models.SmallIntegerField(
+        verbose_name=_("show participant data"),
```
felixrindt, line 58: help text: "Everyone else will only be able to see anonymized data."

ephios/plugins/baseshiftstructures/templates/baseshiftstructures/participation_card_inline.html
felixrindt: We came up with the blur. I'll remove the popover.

jeriox (approved):

PR #1503, opened 2025-03-04
assign workinghours to participants on federated instances

### LINKED ISSUE #1501: Assign workinghours for events on federated instances
As a user participating in an event on another/federated instance, I want the working hours for this event to be credited on my home instance.

jeriox:
closes #1501

### DIFF
--- a/ephios/api/serializers.py
+++ b/ephios/api/serializers.py
@@ -12,11 +12,13 @@ from rest_framework.serializers import ModelSerializer
 from ephios.api.fields import ChoiceDisplayField
 from ephios.core.models import (
     AbstractParticipation,
+    Consequence,
     Event,
     EventType,
     Qualification,
     Shift,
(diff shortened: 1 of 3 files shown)


ephios/core/models/users.py
```diff
@@ -386,20 +387,14 @@ class Meta:
 )
 
 
-class Consequence(Model):
+@dont_log  # as we log the specific models
+class AbstractConsequence(PolymorphicModel):
```
felixrindt, line 391: Currently, the code is not robust against consequence handlers being uninstalled (try removing the youth warden plugin). I think it should be, similar to missing structures and signup flow, it shouldn't crash the application. Maybe setting a custom `_default_manager` would be a simple solution?

felixrindt (commented):
I sent you a patch on Signal so you can look through my hacks during trying this out.

Issue #441, opened 2021-04-19
Models missing from Model Logging


felixrindt:
This is a list of things we found while looking at production usage of model logging

* [ ] some models are not being logged
* [ ] #1545
* [x] german translation "Nichts" often has a bad grammatical fit. Maybe we find another word/solution.
* [x] `__str__` of Qualification Grant isn't translated
* [x] LocalParticipation does not have a verbose name
* [x] When `PlaceholderParticipation` is not registered for modellogging, saving a DispositionForm crashes (couldn't reproduce)
* [x] When creating a federated user, some (changed?) qualifications are in `added` of `m2mfield`, but `current` is displayed for object creation
* [x] rendering QualificationConsequence with deleted event

jeriox:
the Qualification object is not registered for modellogging, maybe we want to change that as well

jeriox:
the Resource object is not registered for modellogging, maybe we want to change that as well

jeriox:
JSON fields are not rendered nicely, e.g. shift structure displays as
`Daten zur Positionierung der Teilnahme in der Schicht-Struktur: {'preferred_unit_path': ''}`

lukasrad02:
The log is only visible to admins or members of a group with "Manage Ephios" permission. I would wish that log entries related to events (sign-ups, confirmations, …) are visible to everybody who can edit/manage that event.

felixrindt:
> JSON fields are not rendered nicely, e.g. shift structure displays as `Daten zur Positionierung der Teilnahme in der Schicht-Struktur: {'preferred_unit_path': ''}`

We could built a custom log recorder that translates the json data into something meaningful, similar to shift configuration. (Does that still work?)

We could seperatly built some generic json rendering in a `<pre>` popup or something like that so it doesn't get shown inside the normal paragraph.

felixrindt:
> The log is only visible to admins or members of a group with "Manage Ephios" permission. I would wish that log entries related to events (sign-ups, confirmations, …) are visible to everybody who can edit/manage that event.

If we do that, it needs to look better. Maybe we can built some sort of table that shows changes better than the Fließtext we currently have.

PR #1540, opened 2025-05-07
Various modellogging fixes

felixrindt:
This PR

- adjusts the m2m save signal to record on `post_add` instead of `pre_add`. Before, when creating an object and subsequently adding items to a m2m-relationship, changes from the last addition weren't recorded in the `current` attribute (as they weren't saved yet). There might still be an issue with removing items, I didn't explore that yet.
- adds/fixes some translations and adds logging for some models, use icon for `None`
- fixes a bug when rendering a qualification consequence for which the event was deleted. That came up when rendering an old log entry, so I included it here.

Refs #441, I identified these distinct topics from the comments until today:

- logging doesnt work with bulk operations
- log could be shown to responsibles of events
- recording of specific json data fields could be improved (e.g. using the derivation recorder)
- Design of the log view could be improved/typografically enhanced
	- better whitespace, tabular layout, proper highlighting
	- shortened long text and show full text on hover, same for `<pre>`-formatted json field data

### DIFF
--- a/ephios/modellogging/templates/modellogging/statement.html
+++ b/ephios/modellogging/templates/modellogging/statement.html
@@ -1,14 +1,15 @@
 {% load logentries %}
 {% load i18n %}
-{% translate "<em>None</em>" as NONE %}
 {{ statement.label }}{% if statement.verb %} {{ statement.verb }}{% endif %}:
 {% if statement.objects %}
     {% for obj in statement.objects %}
         {{ obj|linkify_absolute_url }}{% if not forloop.last %}, {% endif %}
     {% endfor %}
 {% else %}
-    {% if statement.old_value %}
-        {{ statement.old_value|linkify_absolute_url|default_if_none:NONE }} →
-    {% endif %}
-    {{ statement.value|linkify_absolute_url|default_if_none:NONE }}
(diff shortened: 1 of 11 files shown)


ephios/modellogging/templates/modellogging/statement.html
```diff
@@ -1,14 +1,15 @@
...
 {% if statement.objects %}
     {% for obj in statement.objects %}
         {{ obj|linkify_absolute_url }}{% if not forloop.last %}, {% endif %}
     {% endfor %}
 {% else %}
-    {% if statement.old_value %}
-        {{ statement.old_value|linkify_absolute_url|default_if_none:NONE }} →
-    {% endif %}
-    {{ statement.value|linkify_absolute_url|default_if_none:NONE }}
+    {% with NONE='<i class="far fa-minus-square text-body-secondary"></i>' %}
```
jeriox, line 9: i'm not sure if i like the icon/if it is clear what it means. and it at least needs to be aligned with the arrow 
image

jeriox (approved):

ephios/modellogging/templates/modellogging/statement.html
felixrindt: I think visually the whole list isn't good. I can revert the icon change for now until we do a bigger redesign.

PR #1557, opened 2025-05-22
upgrade to Django 5.1

felixrindt:
* [x] dependency upgrade
* [x] implement changes from the framework
* [x] instead of using the new LoginRequiredMiddleware just yet, implement a test checking every view for the use of some form of access management

### DIFF
--- a/tests/core/test_views_auth.py
+++ b/tests/core/test_views_auth.py
@@ -1,3 +1,72 @@
+from django.contrib.auth.mixins import AccessMixin
+from django.urls import get_resolver, reverse
+from guardian.mixins import LoginRequiredMixin
+
+
 def test_accounts_login(django_app):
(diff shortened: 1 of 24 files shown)


tests/core/test_views_auth.py
```diff
@@ -1,3 +1,72 @@
...
     assert resp.status_code == 200, "Should return a 200 status code"
+
+
+def test_accounts_logout(django_app, superuser):
+    resp = django_app.get(reverse("core:oidc_logout"), user=superuser).follow()
+    assert resp.status_code == 200, "Should return a 200 status code"
+    assert not hasattr(resp.request, "user")
+
+
+def test_all_views_are_secured(django_app):
```
jeriox, line 17: this again raises the question how such tests can/should also be run for our plugins

jeriox (approved):

PR #1559, opened 2025-05-24
Add questionnaires

### LINKED ISSUE #1346: Questionaires for shift participants
For some of our shifts, we want to choose participants ourselves. Hence, we're using the "Anfragen und bestätigen" mode for the shifts. To be able to make sensible decisions when accepting participants, we often ask them to specify something in the comment (e.g. previous experience, motivation, …). However, the comment field is only visible if a user clicks the cog icon next to the "Anfragen" button and some of our users do not notice our request to leave a comment in the event description, so we have to email them to ask for the requested information, which takes up additional time and information is scattered across different sources.

For shifts in "Anfragen und bestätigen" mode, I would like to be able to enforce users to leave a comment, i.e., the "Anfragen" button should behave as the cog button currently does. Users may still submit an empty comment, but at least, they have been prompted for input, hopefully reducing the amount of users who do not notice the comment field at all.

For an effective prompt, I would like to be able to specify a custom label/question for the comment field since users who did not notice that we were asking for additional information will most likely ignore a generic "Comment" field.

Additionally, it would be nice if the comment field would be a textarea (larger and allows multi-line input).

- Add a separate mode that behaves as requested above instead of adding configuration options to the existing mode. The new mode might even feature forms/questionnaires (i.e., multiple input fields)
- As a first step, it might make sense to improve the discoverability of the comment field, e.g. by adding a tooltip to the rather generic cog icon or using a labeled button instead of an icon

**Additional context**
*none*

lukasrad02:
Closes #1346


lukasrad02:
Sorry for closing this PR, this was by accident.

The feature is almost complete, what is missing is showing the answers in the disposition. Feel free to take a first look whether the general implementation is fine so far (I will handle things like missing translations later on, so don't spend any time on those details)

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,112 @@
...
+            "help_text": self.description,
+            "initial": initial,
+            "required": required,
+        }
+        if self.type == self.TYPE_SINGLE_RADIO:
+            form_kwargs["widget"] = forms.RadioSelect
+
+        serializer_class = (
+            serializers.CharField
+            if self.type == self.TYPE_TEXT
```
jeriox, line 90: maybe you want to do one if where you set all values related to the type to prevent the multiple nested ifs that are not easy to read

ephios/plugins/questionnaires/signals.py
```diff
@@ -0,0 +1,144 @@
...
+    formfields.append(
+        AdditionalField(
+            "save_answers",
+            forms.BooleanField,
+            {
+                "label": _("Save answers to my user profile for future sign-ups"),
+                "initial": can_save_answers,
+                "disabled": not can_save_answers,
+                "required": False,
+                "widget": forms.HiddenInput if not can_save_answers else forms.CheckboxInput,
```
jeriox, line 93: is there a reason why you still add the field as hidden if it can't be saved anyways?

ephios/plugins/questionnaires/signals.py
```diff
@@ -0,0 +1,144 @@
...
+    **kwargs
+):
+    answers = {}
+
+    for name, value in cleaned_data.items():
+        if name == "questionnaires.save_answers" or not name.startswith("questionnaires."):
+            continue
+
+        # Field name format: questionnaires.123-question-name
+        question_pk = name.split(".", maxsplit=1)[1].split("-")[0]
```
jeriox, line 129: this seems a bit magical if one doesn't read the creation part next to it. maybe you can refactor this to a common spot/helper method?

ephios/plugins/questionnaires/views.py
```diff
@@ -0,0 +1,36 @@
...
+    success_url = reverse_lazy("questionnaires:question_list")
+
+
+class QuestionUpdateView(CustomPermissionRequiredMixin, UpdateView):
+    permission_required = "questionnaires.change_question"
+    form_class = QuestionForm
+    template_name = "questionnaires/question_form.html"
+    success_url = reverse_lazy("questionnaires:question_list")
+
+    def get_queryset(self):
```
jeriox, line 29: i think this is not necessary?

ephios/core/migrations/0036_abstractparticipation_questionnaire_answers_and_more.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 5.0.14 on 2025-05-31 16:37
```
jeriox, line 1: you are missing a migration for the question model

ephios/core/models/users.py
```diff
@@ -117,6 +117,9 @@ class UserProfile(guardian.mixins.GuardianUserMixin, PermissionsMixin, AbstractB
     disabled_notifications = JSONField(
         default=list, encoder=CustomJSONEncoder, decoder=CustomJSONDecoder
     )
+    saved_questionnaire_answers = JSONField(
```
jeriox, line 120: will there be a view to view/(update)/delete saved answers for this PR? if not please create an issue for it

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,112 @@
...
+            initial = None
+
+        field_name = slugify(f"{self.pk} {self.name}")
+        required = self.required if initial is not None else False
+
+        form_class = (
+            forms.CharField
+            if self.type == self.TYPE_TEXT
+            else (
+                forms.MultipleChoiceField if self.type == self.TYPE_MULTIPLE else forms.ChoiceField
```
jeriox, line 76: this renders as a plain HTML multiselect. please use a select2 widget

ephios/plugins/questionnaires/templates/questionnaires/question_form.html
```diff
@@ -0,0 +1,92 @@
...
+        {% csrf_token %}
+        {{ form|crispy }}
+        <div id="choices-formset" class="formset" data-formset data-formset-prefix="{{ form.choices.prefix }}">
+            {{ form.choices.management_form }}
+            <div class="card mb-2">
+                <div class="card-body">
+                    <h5 class="card-title">
+                        {% translate "Choices" %}
+                    </h5>
+                    <p class="text-muted">{% translate "For single- and multiple choice questions" %}</p>
```
jeriox, line 34: please hide this section if it is not applicable

ephios/plugins/questionnaires/signals.py
lukasrad02: I'm reading the value in <https://github.com/lukasrad02/ephios/blob/00988c7c22f49657f9f4b0ef5899566f8aacfe5a/ephios/plugins/questionnaires/signals.py#L135>. By always adding the field, I don't have to handle any special cases when saving the form.

ephios/plugins/questionnaires/signals.py
lukasrad02: Thanks for the hint. Refactored in e69dceeb3f4e05817f9ab47d3b02433f7263d847

ephios/plugins/questionnaires/models.py
lukasrad02: Fixed in e69dceeb3f4e05817f9ab47d3b02433f7263d847

ephios/plugins/questionnaires/templates/questionnaires/question_form.html
lukasrad02: Implemented in 04e070b4300e390c1a96fa6628272f72dde611f7

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,120 @@
...
+        elif (
+            isinstance(participant, LocalUserParticipant)
+            and answer_key in participant.user.saved_questionnaire_answers
+        ):
+            initial = participant.user.saved_questionnaire_answers[answer_key]
+        else:
+            initial = None
+
+        # Reset answer if selected option does not exist
+        if self.type != self.TYPE_TEXT and initial not in self.choices:
```
felixrindt, line 67: Can we make this more extensible (if we ever add number or time questions)? The check would need to look at whether the question type uses choices.

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,120 @@
...
+            serializer_class = serializers.ChoiceField
+            form_kwargs["widget"] = forms.RadioSelect
+        elif self.type == self.TYPE_SINGLE_LIST:
+            form_class = forms.ChoiceField
+            serializer_class = serializers.ChoiceField
+            form_kwargs["widget"] = Select2Widget
+        elif self.type == self.TYPE_MULTIPLE:
+            form_class = forms.MultipleChoiceField
+            serializer_class = serializers.MultipleChoiceField
+            form_kwargs["widget"] = Select2MultipleWidget
```
felixrindt, line 100: Consider using a match statement for this.

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,120 @@
...
+        return AdditionalField(
+            field_name, form_class, form_kwargs, serializer_class, serializer_kwargs
+        )
+
+    def get_form_slug(self):
+        return slugify(f"{self.pk} {self.name}")
+
+    @staticmethod
+    def get_pk_from_slug(slug: str):
+        return slug.split(".", maxsplit=1)[1].split("-")[0]
```
felixrindt, line 120: Please add a docstring with an example.

ephios/core/models/events.py
```diff
@@ -279,6 +279,10 @@ def labels_dict(cls):
     """
     finished = models.BooleanField(default=False, verbose_name=_("finished"))
 
+    questionnaire_answers = models.JSONField(
+        default=dict, verbose_name=_("Answers to questionnaire")
+    )
```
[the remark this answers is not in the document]
lukasrad02: If I recall correctly, you (or Julian) proposed to add fields to the core models in the style of shift structures and signup methods instead of creating separate tables.

In case of separate tables, there should also be a mechanism for plugins to copy their data when a shift is copied (less important for this field on a participation, but especially for `Shift.questionnaire`)

ephios/core/signup/forms.py
```diff
@@ -185,6 +194,22 @@ def __init__(self, *args, **kwargs):
...
+        responses = signup_formfields.send(
+            sender=None,
+            shift=self.shift,
+            participant=self.participant,
+            participation=self.instance,
+            signup_choice=self.data.get("signup_choice"),
+        )
+
+        for _, additional_fields in responses:
+            if isinstance(additional_fields, AdditionalFieldList):
```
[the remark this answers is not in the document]
lukasrad02: Ah, yes, the docstring might be outdated since I actually started with dicts, but this approach seemed brittle to me, I don't like magic strings as identifiers :sweat_smile:.

In case of switching to dicts, would you drop the nesting (`f"{additional_fields.provider_id}.{fieldname}"`)?

ephios/core/signup/forms.py
```diff
@@ -127,12 +131,16 @@ def get_customization_notification_info(self):
 
 
 class BaseSignupForm(BaseParticipationForm):
+    CHOICE_SIGNUP = "sign_up"
+    CHOICE_CUSTOMIZE = "customize"
+    CHOICE_DECLINE = "decline"
```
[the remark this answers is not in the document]
lukasrad02: Thank you, I didn't know about this. Changed in 7743b29e0c447b2f75cc8ffee2344fc441324f65

ephios/plugins/questionnaires/models.py
lukasrad02: Added in 8fa586a3fadb430fd1493e34af729970ea4b73cf

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,120 @@
...
+        max_length=20, verbose_name=_("Type"), choices=TYPE_CHOICES, default=TYPE_TEXT
+    )
+    choices = models.JSONField(default=list, verbose_name=_("Choices"))
+
+    class Meta:
+        verbose_name = _("Question")
+        verbose_name_plural = _("Questions")
+
+    def __str__(self):
+        # pylint: disable=invalid-str-returned
```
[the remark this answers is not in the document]
lukasrad02: Yes, this also works. Changed in bee1d459d45925ca0f185a8ef3597ea3aa88a183

ephios/core/signup/forms.py
```diff
@@ -185,6 +194,22 @@ def __init__(self, *args, **kwargs):
             self.fields["individual_start_time"].disabled = True
             self.fields["individual_end_time"].disabled = True
 
+    def collect_fields(self):
+        responses = signup_formfields.send(
+            sender=None,
+            shift=self.shift,
+            participant=self.participant,
+            participation=self.instance,
```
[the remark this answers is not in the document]
lukasrad02: You mean `self.instance`? As far as I'm aware, Django automatically creates a new instance for model forms, so this should never be `None`. Should I explicitly add this to the docstring although it's the default behavior?

ephios/core/signup/structure/abstract.py
```diff
@@ -49,12 +52,23 @@ def disposition_participation_form_class(self):
         """
         raise NotImplementedError()
 
-    @property
-    def signup_form_class(self):
-        """
-        This form will be used for participations in signup.
-        """
-        raise NotImplementedError()
+    def get_signup_formfields(
```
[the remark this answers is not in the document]
lukasrad02: Renamed this method and some other occurrences of `formfield` in c8734841fcca079a41584b779bfd83f825b3ec13

ephios/plugins/baseshiftstructures/signals.py
```diff
@@ -18,3 +20,29 @@ def register_base_shift_structures(sender, **kwargs):
...
+)
+def provide_signup_formfields(
+    sender,
+    shift: Shift,
+    participant: AbstractParticipant,
+    participation: AbstractParticipation,
+    signup_choice,
+    **kwargs
+):
+    return shift.structure.get_signup_formfields(shift, participant, participation, signup_choice)
```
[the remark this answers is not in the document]
lukasrad02: Right, I just didn't notice that I was crossing an app boundary here. Moved to core in e1f010dea87e3439f1101873470134eee8e7cbf2

ephios/plugins/baseshiftstructures/structure/named_teams.py
```diff
@@ -415,3 +346,73 @@ def get_list_export_data(self):
...
+                not shift.signup_flow.uses_requested_state
+                and not team_stats[team["uuid"]].has_free()
+            ):
+                full_teams.append(team)
+            else:
+                enabled_teams.append(team)
+
+        initial = participation.structure_data.get("preferred_team_uuid")
+        required = (
+            # TODO: This is flawed, as the fields will not be required in customize mode
```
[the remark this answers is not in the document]
jeriox: My bug report is actually not related to this line of code as it has nothing to do with named teams.
to reproduce:
- add a question that is single choice and required. add some options.
- add this question to a shift
- press the signup button for that shift. 
- the participation is created without the user being asked about the question. the participation contains an empty response for the question.

This does not apply for text questions, there the required attribute works correctly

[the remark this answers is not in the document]
felixrindt: > So you consider the current behavior (field is not marked as required until the user submits the form) acceptable?

It is not required when declining, so we don't know when rendering the view. That's why we accepted it. We could mark it required on GET though, as declining via the signup view is probably the less common scenario.

ephios/core/signup/forms.py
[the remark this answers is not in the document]
felixrindt: Nevermind then :)

ephios/core/signup/forms.py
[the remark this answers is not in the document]
felixrindt: magic strings aren't uncommon in this environment, so I'd expect a dict. Also, yes! - no need for nesting if field identifiers are unique, though I would leave that to signal receivers. Like this?
```python
additional_fields_from_all_receivers = {
  "questionaires_12": {
    "form_field": forms.BooleanField,
    ...
  },  
  ...
}
```

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,126 @@
...
+from ephios.core.models.events import AbstractParticipation
+from ephios.core.signup.forms import AdditionalField
+from ephios.core.signup.participants import AbstractParticipant, LocalUserParticipant
+
+
+class Question(models.Model):
+    class Type(models.TextChoices):
+        TEXT = "text", _("Text input")
+        SINGLE_RADIO = "single_radio", _("Single choice")
+        SINGLE_LIST = "single_list", _("Single choice (list)")
```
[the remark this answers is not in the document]
lukasrad02: Implemented in 201704d6efbb41c56c1b18bd2e7975776d0d47aa

ephios/core/models/events.py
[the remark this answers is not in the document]
lukasrad02: Moved this and a user's `saved_questionnaire_answers` to a dedicated model in 59c0ce275544d4e5140465c143e3b0ed1df2de41

ephios/core/models/events.py
```diff
@@ -369,6 +373,12 @@ class Shift(DatetimeDisplayMixin, Model):
         decoder=CustomJSONDecoder,
     )
 
+    questionnaire = JSONField(
+        default=list,
+        encoder=CustomJSONEncoder,
+        decoder=CustomJSONDecoder,
+    )
```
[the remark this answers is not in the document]
lukasrad02: Moved to a dedicated model and added a `shift_copy` signal in e14133bc6b47095edde960cd7b585721d197ec66

ephios/core/signup/forms.py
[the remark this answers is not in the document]
lukasrad02: Changed to dicts in 306e1b3c5f21934f7a091f4a71ac06cfe76a27a6

ephios/plugins/questionnaires/forms.py
```diff
@@ -3,11 +3,15 @@
...
-from ephios.plugins.questionnaires.models import Question
+from ephios.plugins.questionnaires.models import Answer, Question
 
 
 class ChoiceForm(forms.Form):
-    name = forms.CharField(label=_("Choice name"))
+    name = forms.CharField(
+        label=_("Choice name"),
+        # pylint: disable=protected-access
+        max_length=Answer._meta.get_field("answer").max_length,
```
felixrindt, line 13: I never thought about this before, but are the lengths considered the same for 2+ byte unicode chars in forms and the database? The database cares about bytes, but the for might care about actual character count, no?

lukasrad02: According to <https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.CharField.max_length>, `max_length` on the model field also refers to the number of *characters*, so this should not be an issue

ephios/plugins/baseshiftstructures/structure/named_teams.py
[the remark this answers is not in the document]
jeriox: Could you reproduce with my description?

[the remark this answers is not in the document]
lukasrad02: Yes, thanks for the reminder. Should be fixed in 7e488f63277049ebe2dbade6e778bcb00e4231f2

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,162 @@
...
+    participation = models.ForeignKey(AbstractParticipation, on_delete=models.CASCADE)
+    question = models.ForeignKey(Question, on_delete=models.PROTECT)
+    answer = models.CharField(max_length=100, verbose_name=_("Answer"))
+
+    class Meta:
+        verbose_name = _("Answer")
+        verbose_name_plural = _("Answers")
+
+    def __str__(self):
+        return str(self.answer)
```
jeriox, line 149: please use more context (question and participation). This is useful when registering the models for the admin view (which you should please do by adding the following to admin.py in the questionnaires folder):

```
from django.contrib import admin

from ephios.plugins.questionnaires.models import Questionnaire, Question, Answer, SavedAnswer

admin.site.register(Questionnaire)
admin.site.register(Question)
admin.site.register(Answer)
admin.site.register(SavedAnswer)
``

File: ephios/plugins/questionnaires/models.py [hunk 24]
```diff
@@ -0,0 +1,162 @@
...
+        field["form_kwargs"] = {
+            "label": self.question_text,
+            "help_text": self.description,
+            "initial": initial,
+            "required": required,
+        }
+
+        field["serializer_kwargs"] = {
+            "required": required,
+        }
```
felixrindt, line 84: can you put this into the dict literal in lines 70/72 please?

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,162 @@
...
+                field["form_class"] = forms.MultipleChoiceField
+                field["serializer_class"] = serializers.MultipleChoiceField
+                field["form_kwargs"]["widget"] = Select2MultipleWidget
+
+        if self.type != self.Type.TEXT:
+            assert isinstance(
+                self.choices, list
+            ), f"The choices of question {self.name} are not a list"
+            # pylint: disable=not-an-iterable
+            choices = [(choice, choice) for choice in self.choices]
```
felixrindt, line 110: this looks dangerous, can every arbitrary string be a choice key in HTML selects? What if it contains special characters?

ephios/plugins/questionnaires/models.py
lukasrad02: Changed in 8e272dbcf39e850e08428f5a5bef47bddf18e723

ephios/core/signup/views.py
```diff
@@ -79,6 +86,13 @@ def customize_pressed(self, form):
     def signup_pressed(self, form):
         try:
             participation = form.save()
+            signup_save.send(
```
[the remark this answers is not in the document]
lukasrad02: I initially considered it not necessary, as for example, the shift structures also don't care about decline. But to make the signal more generally usable, it is now (381f05f27125819c4ef0f57f6934c48c604a8428) also sent on decline.

I also delete answers when users decline to make the answer view simpler (can show all answers and don't have to check whether user declined), but for the shift structures, I didn't change the behavior for now. Shall I also remove their stored information?

ephios/plugins/complexsignup/structure.py
```diff
@@ -305,6 +266,60 @@ def get_list_export_data(self):
...
+            # TODO: This is flawed, as the fields will not be required in customize mode
+            signup_choice == SignupForm.SignupChoices.SIGNUP
+            and shift.structure.configuration.choose_preferred_unit
+        )
+
+        complex_structure = self.shift.structure
+        complex_structure._assume_cache()
+        choices = [
+            (b["path"], _build_human_path(b))
+            for b in atomic_block_participant_qualifies_for(
```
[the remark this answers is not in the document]
lukasrad02: Changed in 80ca01a03379ce69c7bcdee94b3f58e1bf4b53f7.

Thank you for the warning on merge conflicts

ephios/plugins/questionnaires/templates/questionnaires/question_list.html
```diff
@@ -0,0 +1,39 @@
...
+    {% translate "Questions" %}
+{% endblock %}
+
+{% block content %}
+    <h1>{% translate "Questions" %}</h1>
+    {% if perms.questionnaires.add_question %}
+        <a class="btn btn-primary text-nowrap" href="{% url "questionnaires:question_add" %}"><span
+            class="fa fa-plus"></span> <span>{% translate "Add question" %}</span></a>
+    {% endif %}
+    <table class="table table-striped">
```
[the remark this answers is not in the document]
lukasrad02: Changed to bootstrap grid in 5211e9b57d5ba20b0baf3fe6623f67829b2557ab.

I'm not fully satisfied by the pencil icon I chose as icon for the question type, do you have a better idea?

image

ephios/core/signup/views.py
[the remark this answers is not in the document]
felixrindt: I think we should never delete the data a user provided if they change their mind or availability changes. I'd rather see participations filtered for states in places where the data is used.

ephios/plugins/questionnaires/templates/questionnaires/question_list.html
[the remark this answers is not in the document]
lukasrad02: Oh, good idea to actually use different icons. However, I'm afraid that users without technical experience might not understand the icons. And the `check-square` could also be confused with a (not yet existing) "Yes/No question" type.

What do you think about `list` and `tasks`?

image

image

ephios/plugins/questionnaires/models.py
```diff
@@ -46,81 +45,118 @@ def get_signup_form_field(
...
+        verbose_name = _("Answer")
+        verbose_name_plural = _("Answers")
+
+    def __str__(self):
+        return str(self.answer)
+
+
+class SavedAnswer(models.Model):
+    user = models.ForeignKey(UserProfile, on_delete=models.CASCADE)
+    question = models.ForeignKey(Question, on_delete=models.PROTECT)
```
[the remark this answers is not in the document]
lukasrad02: Fixed in 738971b82785e2b9c0e4568d8602059dcd7de584

ephios/plugins/questionnaires/models.py
lukasrad02: Added comment in 15619e09bb9d5df943f24692cd0c3de89bfde34f

ephios/plugins/questionnaires/signals.py
jeriox: you could also just check whether `save_answers in cleaned_data` to make it simpler, but its totally fine this way as well

ephios/plugins/questionnaires/templates/questionnaires/question_list.html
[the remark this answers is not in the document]
lukasrad02: Updated to use different icons per type in 65bdad6cb56f57473c8e42cc6da20fafa1671566

ephios/core/models/users.py
lukasrad02: Implemented in 1fddf194b7b6db7f3433c363926c6a3dd21fda1f

ephios/plugins/complexsignup/structure.py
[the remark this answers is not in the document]
felixrindt: It was merged, github is indicating conflicts.

ephios/core/signals.py
```diff
@@ -101,6 +101,15 @@
...
 
+shift_action = PluginSignal()
+"""
+This signal is sent out to collect additional actions that managers can perform on on a shift. For
+each action, a button will be displayed in the shift card next to the disposition button. Receivers
+of the signal will receive the ``shift`` and ``request`` and are expected to return an array of
+``{label: str, url: str}`` dicts, representing the available actions.
+The buttons will only be shown to managers of the respective shift.
+"""
+
```
felixrindt, line 112: We should be careful with the naming here. "event_action" refers to the dropdown menu on event, so I'd expect "shift_action" to also be about the dropdown menu. I think renaming `event_action` to `event_menu` is the better way out though.

lukasrad02: Naming-wise, I agree that I prefer renaming the event-related signal.

However, this would be a breaking change. Is that okay?

felixrindt: We will have breaking signal renaming in the next release any way (due to #1606), so lets do it now. There aren't many plugins we have that use the signal. Other people will get the error. This is why we are still not at version 1. ^^

ephios/plugins/complexsignup/structure.py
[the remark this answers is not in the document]
lukasrad02: Conflicts resolved during rebase

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,214 @@
...
+class Questionnaire(models.Model):
+    shift = models.OneToOneField(Shift, on_delete=models.CASCADE)
+    questions = models.ManyToManyField(Question, blank=True)
+
+    class Meta:
+        verbose_name = _("Questionnaire")
+        verbose_name_plural = _("Questionnaires")
+
+    def __str__(self):
+        return f"{", ".join(self.questions.values_list("name", flat=True))} @ {self.shift}"
```
felixrindt, line 188: broken in CI: beware that before python3.12, you can't mix quotation styles in f-strings and their substitution fields:


```suggestion
        return f'{", ".join(self.questions.values_list("name", flat=True))} @ {self.shift}'
```

ephios/core/signals.py
```diff
@@ -103,6 +105,21 @@
 """
 
 
+shift_action = PluginSignal()
+"""
+This signal is sent out to collect additional actions that managers can perform on on a shift. For
+each action, a button will be displayed in the shift card next to the disposition button. Receivers
+of the signal will receive the ``shift`` and ``request`` and are expected to return an array of
+``{label: str, url: str}`` dicts, representing the available actions.
+The buttons will only be shown to managers of the respective shift.
```
jeriox, line 114:
```suggestion
The buttons will only be shown to responsibles of the respective shift.
```

ephios/core/templates/core/disposition/fragment_participation.html
```diff
@@ -43,6 +46,9 @@ <h6 class="d-inline">
             <button class="btn" type="button" data-bs-toggle="collapse"
                     data-bs-target="#participation-collapse-{{ form.instance.id }}" aria-expanded="false"
                     aria-controls="collapseExample">
+                {% if additional_html %}
+                    <span class="fas fa-info-circle pe-2"></span>
```
jeriox, line 50: I'm not sure if it is smart to do highlighting just if additional html is present.
- There may be a usecase where a plugin wants to render HTML for every participation which is not critical for disposition, but nice-to-have context
- When a question is required, every participation will have this indicator.

Both could lead to a visual overload for the user without providing a benefit/additional information

ephios/plugins/questionnaires/signals.py
```diff
@@ -174,3 +178,30 @@ def save_signup(
...
+        "questionnaires/disposition_participation_answers.html", {"answers": answers}
+    )
+
+
+@receiver(shift_action, dispatch_uid="ephios.plugins.questionnaires.signals.shift_action")
+def add_shift_action(sender, shift: Shift, request, **kwargs):
+    if hasattr(shift, "questionnaire") and shift.questionnaire and shift.questionnaire.questions:
+        return [
+            {
+                "label": "View answers",
```
jeriox, line 204: is it clear for the user what "answers" is referring to? But mentioning questionnaires here would probably lead to a button that is too big

ephios/plugins/questionnaires/templates/questionnaires/disposition_participation_answers.html
```diff
@@ -0,0 +1,15 @@
+{% load questionnaires_extras %}
+
+<h4>Questionnaire Answers</h4>
```
jeriox, line 3: i think a smaller heading would do

ephios/plugins/questionnaires/templates/questionnaires/aggregate_answers.html
```diff
@@ -0,0 +1,65 @@
...
+
+{% block title %}
+    {% blocktrans trimmed %}Answers for {{ shift }}{% endblocktrans %}
+{% endblock %}
+
+{% block content %}
+    <h1>{% blocktrans trimmed %}Answers for {{ shift }}{% endblocktrans %}</h1>
+
+    {% trans "This page only shows answers of confirmed participants. You can see the answers of other participants in the disposition view." as alert %}
+    {% render_alert alert "warning" False %}
```
jeriox, line 14: maybe info is enough as a level?

ephios/plugins/questionnaires/templates/questionnaires/savedanswer_form.html
```diff
@@ -0,0 +1,32 @@
...
+{% block title %}
+    {% translate "Edit saved questionnaire answer" %}
+{% endblock %}
+
+{% block settings_content %}
+    <h3>
+        {% translate "Edit saved questionnaire answer" %}
+    </h3>
+
+    <h4>
```
jeriox, line 14: this looks a bit weird with the two heading directly below each other

ephios/plugins/questionnaires/templates/questionnaires/savedanswer_form.html
```diff
@@ -0,0 +1,32 @@
...
+        {{ savedanswer.question.question_text }}
+    </h4>
+
+    {% if savedanswer.question.description %}
+        <p>
+            {{ savedanswer.question.description }}
+        </p>
+    {% endif %}
+
+    <form method="post">
```
jeriox, line 24: maybe we should display a warning that this does not change the answer for already existing participations?

ephios/plugins/questionnaires/templates/questionnaires/savedanswer_list.html
```diff
@@ -0,0 +1,42 @@
...
+                            class="fa fa-edit"></span> <span
+                                class="d-none d-md-inline">{% translate "Edit" %}</span></a>
+                        <a class="btn btn-secondary btn-sm text-nowrap ms-1"
+                           href="{% url "questionnaires:saved_answers_delete" answer.question.pk %}"><span
+                            class="fa fa-trash-alt"></span> <span
+                                class="d-none d-lg-inline">{% translate "Delete" %}</span></a>
+                    </div>
+                </div>
+            </li>
+        {% endfor %}
```
jeriox, line 40: please add a message for empty

ephios/plugins/questionnaires/templates/questionnaires/savedanswer_list.html
```diff
@@ -0,0 +1,42 @@
...
+                        <h5 class="mb-0 d-inline">
+                            {{ answer.question.name }}
+                        </h5>
+                        <br/>
+                        <span class="text-muted">{{ answer.question.question_text }}</span>
+                    </div>
+                    <div class="col-12 col-lg-5">
+                        {{ answer.answer | answer_text }}
+                    </div>
+                    <div class="col-12 col-lg-auto d-flex align-items-end align-items-lg-center justify-content-end">
```
jeriox, line 28: i think the buttons could go to the right side on large screens, the space seems a bit wasted atm

ephios/plugins/questionnaires/templates/questionnaires/disposition_participation_answers.html
lukasrad02: Oh, I think the size of the other labels (individual start time etc) in the card has changed with the rebase. In the new layout, I agree that this heading is to large.

Changed in ca8906a972719757cce4f2b3901bb6c3f16f247b

ephios/plugins/questionnaires/templates/questionnaires/savedanswer_form.html
lukasrad02: Removed the generic heading in 030d358c68f7d6dc8ffa1e7297760e6d5dea9f7d

ephios/plugins/questionnaires/templates/questionnaires/savedanswer_form.html
lukasrad02: Added in dced00e1ba1b988f93ff581c9fedcb6b6f006c34

ephios/plugins/questionnaires/templates/questionnaires/savedanswer_list.html
lukasrad02: Added in dd04c0d97c56b571df28c0905edeefbe214cfaca, also for the question list

ephios/plugins/questionnaires/templates/questionnaires/aggregate_answers.html
```diff
@@ -0,0 +1,65 @@
...
+                            <tr>
+                                <td>{{ answer }}</td>
+                                <td>{{ count }}</td>
+                            </tr>
+                        {% endfor %}
+                    </tbody>
+                </table>
+            {% else %}
+                <p>
+                    <em>{% trans "No answers so far" %}</em>
```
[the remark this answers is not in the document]
lukasrad02: Thanks for pointing this out. Changed in 880a4dd431cfffe508902090e18ce55d6e48899e

ephios/plugins/questionnaires/templates/questionnaires/savedanswer_list.html
lukasrad02: What do you mean? On a sufficiently large screen, the buttons are on the right:

image

Reducing the breakpoint from `lg` to `md` does not help, since the column of the buttons is to wide to fit on the right, so there is still a line break. And I would rather not make the columns for the question and answer smaller, since both texts may be longer than just a few characters…

ephios/core/templatetags/event_extras.py
```diff
@@ -149,6 +149,16 @@ def event_plugin_actions(event_detail_view):
     return mark_safe(html)
 
 
+@register.simple_tag(name="shift_plugin_actions")
+def shift_plugin_actions(shift, request):
+    return [
+        action
+        for _, actions in shift_action.send(None, shift=shift, request=request)
+        if actions
```
felixrindt, line 157: I did not know this syntax works

ephios/core/signup/structure/abstract.py
```diff
@@ -49,12 +52,22 @@ def disposition_participation_form_class(self):
...
+        signup_choice,
+    ):
+        pass
+
+    def save_signup(
+        self,
+        shift: Shift,
+        participant: AbstractParticipant,
+        participation: AbstractParticipation,
+        cleaned_data: dict[str, Any],
```
felixrindt, line 68: cleaned data does contain signup_choice, right? I think it might be nice to explicitly have it as kwargs so this methods signature is an extension of get_signup_form_fields

ephios/plugins/questionnaires/urls.py
```diff
@@ -0,0 +1,72 @@
...
+        QuestionListView.as_view(),
+        name="question_list",
+    ),
+    path(
+        "questions/archived/",
+        QuestionListView.as_view(show_archived=True),
+        name="question_list_archived",
+    ),
+    path(
+        "questions/add/",
```
felixrindt, line 28: damn I never noticed. Sometimes we add, sometimes we create. I vote for "create" everywhere.

felixrindt (commented):
Wow, most of this stuff looks like out of a Django Textbook!

ephios/plugins/questionnaires/urls.py
```diff
@@ -0,0 +1,72 @@
...
+        QuestionCreateView.as_view(),
+        name="question_add",
+    ),
+    path(
+        "questions/<int:pk>/edit/",
```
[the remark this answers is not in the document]
lukasrad02: Created #1613 to track updating other URLs. Nothing to update for this PR.

ephios/plugins/questionnaires/models.py
```diff
@@ -0,0 +1,241 @@
...
+    ModelFieldsLogConfig(
+        attach_to_func=lambda answer: (AbstractParticipation, answer.participation_id)
+    ),
+)
+
+
+class SavedAnswer(models.Model):
+    user = models.ForeignKey(UserProfile, on_delete=models.CASCADE)
+    question = models.ForeignKey(Question, on_delete=models.CASCADE)
+    answer = models.JSONField(verbose_name=_("Answer"))
```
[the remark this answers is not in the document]
lukasrad02: This is due to the form field being generated manually to use the right field and widget type and choices (if applicable).

Fixed in 3bf3defa1a077f11028688057c33c5964a0d17b9

ephios/plugins/questionnaires/templates/questionnaires/disposition_participation_answers.html
```diff
@@ -0,0 +1,15 @@
+{% load questionnaires_extras %}
+
+<h5>Questionnaire Answers</h5>
+
+<ul>
+    {% for answer in answers %}
+        <li>
```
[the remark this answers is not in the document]
felixrindt: Hmm maybe we use the flush group elsewhere? I do like it somehow.

ephios/plugins/questionnaires/templates/questionnaires/question_list.html
```diff
@@ -0,0 +1,103 @@
...
+                        {% endif %}
+                    </div>
+                    <div class="col-4 col-lg-auto d-flex align-items-end align-items-lg-center justify-content-end">
+                        {% if show_archived %}
+                            <a class="btn btn-secondary btn-sm text-nowrap ms-1"
+                               href="{% url "questionnaires:question_unarchive" question.pk %}"><span
+                                class="fa fa-undo"></span> <span
+                                    class="d-none d-md-inline">{% translate "Unarchive" %}</span></a>
+                        {% else %}
+                            <a class="btn btn-secondary btn-sm text-nowrap ms-1"
```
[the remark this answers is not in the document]
felixrindt: I do find responsive layouting very tricky and spent a lot of time tinkering with the event list. Don't worry about it too much in this view, even if it's not the most compact layout in the end.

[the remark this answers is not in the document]
lukasrad02: Okay, then I will stick with my current solution for now

ephios/plugins/questionnaires/templates/questionnaires/disposition_participation_answers.html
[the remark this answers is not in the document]
lukasrad02: Okay, then we can keep the implementation from 556c467780a7f8849eb9467d26c38f32eb7cee47 without any further changes

ephios/plugins/complexsignup/structure.py
```diff
@@ -373,6 +331,55 @@ def get_list_export_data(self):
                 )
         return export_data
 
+    def get_signup_form_fields(self, participant, participation, signup_choice):
```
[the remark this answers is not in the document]
lukasrad02: You mean extracting this method to some strategy-pattern-like class that the shift structure can reference?

I don't think that this would be an improvement, as there will hardly be any case where different shift structures will share the same strategy/FormField. On the other hand, shift structures would have to expose more of their internals towards the FormField (such as `_assume_cache` for the complex structure or `_get_signup_stats_per_group` for the named teams).  
So I'm afraid it would just be some additional indirection…

ephios/plugins/questionnaires/templatetags/questionnaires_extras.py
```diff
@@ -0,0 +1,45 @@
+from django import template
+from django.utils.translation import gettext_lazy as _
+
+from ephios.core.models.events import AbstractParticipation, Shift
+from ephios.plugins.questionnaires.models import Answer, Question
+
+register = template.Library()
```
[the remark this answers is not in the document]
lukasrad02: Thanks for the pointers for the different tags and filters. I've integrated all of them into the correct locations in a46debda38f455237719f3585f481667fb13a979

ephios/plugins/complexsignup/structure.py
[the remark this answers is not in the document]
felixrindt: not a strategy class, rather a highly specialized FormField. This is not about reusability. What bugs me is that inside the function `get_signup_form_fields` there are a lot of lines configuring a single field. It does not have the "getter" vibes. If you don't like the idea feel free to close this discussion, it's everything but critical.

ephios/plugins/questionnaires/templates/questionnaires/question_form.html
```diff
@@ -0,0 +1,100 @@
...
+                {% endescapescript %}
+            </script>
+        </div>
+        <div class="form-group">
+            <a class="btn btn-secondary" href="{% url "questionnaires:question_list" %}">{% translate "Cancel" %}</a>
+            <input type="submit" class="btn btn-primary float-end" value="{% translate "Save" %}">
+        </div>
+    </form>
+
+    <script nonce="{{ request.csp_nonce }}" type="text/javascript" src="{% static "questionnaires/choice_visibility.js" %}"></script>
```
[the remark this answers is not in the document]
felixrindt: perfect! Thanks!

ephios/plugins/questionnaires/forms.py
```diff
@@ -0,0 +1,107 @@
...
+            initial=[{"name": choice} for choice in self.instance.choices],
+            data=data,
+            prefix="choices",
+        )
+
+    def save(self, commit=True):
+        self.instance.choices = [
+            choice["name"]
+            for choice in self.choices.cleaned_data
+            if "name" in choice and not choice["DELETE"]
```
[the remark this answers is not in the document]
felixrindt: Redundancy as in `choice["name"]` as a comprehension value and `"name" in choice` as the check in the if.

ephios/plugins/questionnaires/forms.py
```diff
@@ -32,9 +32,9 @@ def __init__(self, data=None, **kwargs):
 
     def save(self, commit=True):
         self.instance.choices = [
-            choice["name"]
-            for choice in self.choices.cleaned_data
-            if "name" in choice and not choice["DELETE"]
+            choice_form_data["name"]
+            for choice_form_data in self.choices.cleaned_data
+            if choice_form_data and not choice_form_data["DELETE"]
```
felixrindt, line 37: Now I don't get this check. Why would `choice_form_data` ever be falsy? Doesn't it always contain at least the `DELETE` key?

ephios/plugins/questionnaires/templates/questionnaires/aggregate_answers.html
```diff
@@ -13,29 +12,21 @@ <h1>{% blocktranslate trimmed %}Answers for {{ shift }}{% endblocktranslate %}</
...
     {% render_alert alert "info" False %}
 
-    {% for question in shift.questionnaire.questions.all %}
-        <h2>{{ question.question_text }}</h2>
+    {% for question in question_data %}
+        <h2>{{ question.question.question_text }}</h2>
 
         <p>
-            <em>{{ question.get_type_display }}, {{ question|required_text }}</em>
+            <em>{{ question.question.get_type_display }}, {{ question.question.required|yesno:_("required,optional") }}</em>
```
felixrindt, line 19: huh, today I learned about `_()` in templates

felixrindt (approved):

ephios/core/templates/core/disposition/fragment_participation.html
jeriox: hmmm, difficult decision. I think I am leaning a bit towards not showing anything, as it might be a question where only the aggregation is relevant (e.g. food) or the planners will remember that they specifically added a question that they want to see the answer of.

@felixrindt what is your take on this?

ephios/plugins/questionnaires/templates/questionnaires/savedanswer_list.html
jeriox: ah nevermind this seems to have changed with a more recent commit

ephios/plugins/questionnaires/templates/questionnaires/question_confirm_archive.html
```diff
@@ -0,0 +1,41 @@
...
+    </div>
+
+    <form method="post">
+        {% csrf_token %}
+        {{ form|crispy }}
+        <p>{% blocktranslate trimmed with action=action %}
+            Are you sure you want to {{ action }} the question {{ question }}?
+        {% endblocktranslate %}</p>
+        <a role="button" class="btn btn-secondary" href="{{ back_url }}">{% translate "Back" %}</a>
+        <button type="submit" class="btn btn-danger">{{ action | capfirst }}</button>
```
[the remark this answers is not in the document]
lukasrad02: Changed to use `btn-primary` in this case in 8a19b1a124a62d3e8b758f92fd5a257b02d9c7b9

ephios/plugins/questionnaires/templates/questionnaires/question_list.html
```diff
@@ -39,28 +65,53 @@ <h5 class="mb-0 d-inline">
...
+                                class="fa fa-edit"></span> <span
+                                    class="d-none d-md-inline">{% translate "Edit" %}</span></a>
+                            <a class="btn btn-secondary btn-sm text-nowrap ms-1"
+                               href="{% url "questionnaires:question_archive" question.pk %}"><span
+                                class="fa fa-archive"></span> <span
+                                    class="d-none d-md-inline">{% translate "Archive" %}</span></a>
+                            <a class="btn btn-secondary btn-sm text-nowrap ms-1"
+                               href="{% url "questionnaires:question_delete" question.pk %}"><span
+                                class="fa fa-trash-alt"></span> <span
+                                    class="d-none d-lg-inline">{% translate "Delete" %}</span></a>
```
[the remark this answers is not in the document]
lukasrad02: If the question cannot be deleted, I'm now hiding the button and show a tooltip to explain (8a19b1a124a62d3e8b758f92fd5a257b02d9c7b9)

ephios/plugins/baseshiftstructures/structure/named_teams.py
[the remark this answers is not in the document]
lukasrad02: Finally got some time to revisit this.

> It is not required when declining, so we don't know when rendering the view. That's why we accepted it. We could mark it required on GET though, as declining via the signup view is probably the less common scenario.

Actually, I wasn't even aware that there is a "Decline" button on the customize page.

Since this page is getting even more form fields with this PR, I would also prefer to actually mark all required fields as such on the GET request, so that the users know which of the (possibly many) fields they have to fill.

Do you think it would make sense to remove the Decline button from the page? By implementing this change, it could become unusable in many cases and would leave users with a potentially confusing error message

ephios/core/templates/core/disposition/fragment_participation.html
felixrindt: I'd go with no icon, as in many situations noone or everyone would get one. Let's see how people use the questions, first.

ephios/plugins/baseshiftstructures/structure/named_teams.py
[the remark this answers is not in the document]
lukasrad02: > I think having the fields required by default and turning that off if the signup action is "decline" is a sensible way to go

I thought this wouldn't work due to the client-side validation performed by the browser, but double-checked and learned about the `[formnovalidate](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#formnovalidate)` attribute to override form validation per button.

→ Implemented in 11592bace88aab67503dd0bb2cafa63ca7426c3b

ephios/core/signup/views.py
```diff
@@ -38,94 +41,167 @@ def get_form_kwargs(self):
...
             )
-        return redirect(self.participant.reverse_event_detail(self.shift.event))
+            return None
+
+        for fieldname, field in collect_signup_form_fields(
+            self.shift, self.participant, self.participation, signup_choice
+        ):
+            if field["required"] and not field["default"]:
+                # not supported by quick action
+                messages.info(self.request, _("Please fill out the form to sign up."))
```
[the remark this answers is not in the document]
lukasrad02: Changed message in 3cb318520225f9aa0c5434080bba94b6a8d90146

lukasrad02:
With the tests added in fa1270425636a95c0b62c629c0aa49a401b8a430 an a rebase on main, this PR should finally be ready to be merged!

ephios/core/signup/views.py
```diff
@@ -38,94 +41,170 @@ def get_form_kwargs(self):
...
+            case SignupForm.SignupChoices.CUSTOMIZE if validator.can_customize_signup():
+                # pylint: disable-next=unnecessary-lambda-assignment
+                flow_action = lambda **kwargs: None  # noop
+                error_message = _("There was an error saving your participation.")
+                success_message = _("Your participation was saved.")
+            case _:
+                transaction.rollback()
+                messages.error(self.request, _("This action is not allowed."))
+                # always jump back to event-detail, in case SignupForm should be unreachable
+                return redirect(self.participant.reverse_event_detail(self.shift.event))
```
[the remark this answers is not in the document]
lukasrad02: Yes, this is from your code. But it seems like you have also already pushed a fix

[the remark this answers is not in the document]
lukasrad02: Will the transaction now be actually rolled back although the error is handled?

PR #1579, opened 2025-06-17
upgrade django-guardian to 3.0

felixrindt:
As the update is mostly considered a drop-in replacement (with some changes in how settings are configured), There wasn't really any code I could remove:

- we still need our own PermissionRequiredMixin, as the guardian one didn't add the possibility for redirecting anonymous users to login while raising 403 for authenticated users on lack of permission
- Looks like there will be some new functionality in 3.1: https://github.com/django-guardian/django-guardian/pull/855

### DIFF
--- a/ephios/settings.py
+++ b/ephios/settings.py
@@ -347,7 +347,6 @@ def GET_USERCONTENT_QUOTA():
 
 # Guardian configuration
 ANONYMOUS_USER_NAME = None
-GUARDIAN_MONKEY_PATCH = False
 
 # django-select2
 # Prevent django-select from loading the select2 resources as we want to serve them locally
--- a/poetry.lock
+++ b/poetry.lock
@@ -1060,18 +1060,18 @@ django-jquery-js = "*"
 
 [[package]]
 name = "django-guardian"
-version = "2.4.0"
-description = "Implementation of per object permissions for Django."
+version = "3.0.0"
+description = "Per object permissions for Django"
 optional = false
-python-versions = ">=3.5"
+python-versions = ">=3.9"
 groups = ["main"]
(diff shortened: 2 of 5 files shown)


ephios/settings.py
```diff
@@ -347,7 +347,6 @@ def GET_USERCONTENT_QUOTA():
 
 # Guardian configuration
 ANONYMOUS_USER_NAME = None
-GUARDIAN_MONKEY_PATCH = False
```
jeriox, line 350: why was this here and why don't we need it anymore?

jeriox (approved):
regarding the anon redirect stuff: did we open an issue with guardian/is there one? upstreaming our solution would be nice at some point

poetry.lock
```diff
@@ -1054,18 +1042,18 @@ django-jquery-js = "*"
 
 [[package]]
 name = "django-guardian"
-version = "2.4.0"
-description = "Implementation of per object permissions for Django."
+version = "3.0.0"
```
jeriox, line 1045: why not 3.0.1?

felixrindt: Wasn't available in pypi when I did the upgrade

ephios/settings.py
felixrindt: They renamed it, I reread the docs and I think we can reintroduce it under the new name. 
It disables monkeypatching the default User model, which we don't use, and which didn't cause any issues.

felixrindt:
> regarding the anon redirect stuff: did we open an issue with guardian/is there one? upstreaming our solution would be nice at some point

They do have proposals over there in https://github.com/django-guardian/django-guardian/issues/200

PR #1581, opened 2025-06-17
add status value to ical event feed 

felixrindt:


ephios/core/ical.py
```diff
@@ -58,9 +58,21 @@ def item_start_datetime(self, item):
     def item_end_datetime(self, item):
         return item.participations.all()[0].end_time
 
+    def item_status(self, item):
+        # The ical status field can be CONFIRMED|TENTATIVE|CANCELLED
+        return {
+            AbstractParticipation.States.CONFIRMED: "CONFIRMED",
+            AbstractParticipation.States.REQUESTED: "TENTATIVE",  # displayed less opaque in many calendars
+            AbstractParticipation.States.RESPONSIBLE_REJECTED: "CANCELLED",  # often displayed struck out
```
jeriox, line 66: do we really want to display these as well? usually there is no way of editing imported ICS feeds, and I personally hate it when I have events in my calender that I will not attend

ephios/core/templates/core/settings/settings_calendar.html
```diff
@@ -19,6 +24,14 @@ <h3 class="mt-3">{% translate "Calendar" %}</h3>
                 title="Copied!">{% translate "Copy" %}</button>
     </div>
 
+    <form method="post">
+        {% csrf_token %}
+        <button type="submit" name="reset" value="1" class="btn btn-danger">
+            {% translate "Reset Calendar-URL" %}
```
jeriox, line 30:
```suggestion
            {% translate "Reset calendar URL" %}
```

ephios/core/ical.py
jeriox: Outlook removes the event completely when you cancel an invitation/event. Which clients did you test?

felixrindt: image
This is the nextcloud calendar.

----

Google maps doesn't do anything to it:
image

----

Thunderbird is like nextcloud:
image

---
My FOSS calendar app doesn't display anything either:
image

ephios/core/views/accounts.py
```diff
@@ -278,6 +278,7 @@ def setup(self, request, *args, **kwargs):
     def post(self, request, *args, **kwargs):
         if request.POST.get("confirm"):
             self.object.set_unusable_password()
+            self.object.reset_calendar_token()
```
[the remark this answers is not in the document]
felixrindt: this is not the users password reset code, but rather where managers can reset revoke all use-associated secrets.

PR #1593, opened 2025-07-11
fix assigning designated participants to positions they dont qualify for

felixrindt:
I want to make it all faster too, but I don't know how, really. :/

### DIFF
--- a/ephios/core/signup/disposition.py
+++ b/ephios/core/signup/disposition.py
@@ -36,12 +36,13 @@ class BaseDispositionParticipationForm(BaseParticipationForm):
         label=_("Hide comment for participant"), required=False, label_suffix=""
     )
 
-    def __init__(self, **kwargs):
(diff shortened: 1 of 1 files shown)


ephios/core/services/matching.py
```diff
@@ -120,35 +125,50 @@ def score_pairing(
...
     if participant in confirmed_participants:
         score += confirmed_value
-    score += position.skill_level * max_skill_value
+
+    if is_qualified:
+        # if qualified, lets prefer high skill positions
+        score += position.skill_level * max_skill_value
+    else:
+        # if not qualified (but designated), let's prefer low skill positions
+        score -= position.skill_level * max_skill_value
```
felixrindt, line 170: this is new, and led to designated participants being put to *any* position of the designated atomic block, because being qualified wasn't taken into account in that case.

PR #1598, opened 2025-07-17
colorful own participation pills

felixrindt:
Could help find users their own participation at a glance. Could use other ways of course, I just looked at styling options in [the bootstrap docs](https://getbootstrap.com/docs/5.3/components/badge/)...

image


jeriox:
I like the idea. However, I'm not sure about adding more bright colors in there, especially because yellow is already used to flag minor participants here.

On the other hand, I just asked myself what the purpose of this feature is, and in my mind it would be to answer the question "which unit am i dispatched to?". For that, a more decent highlighting (e.g. an outline around the participation) would be enough IMO. The state of the own participation is already communicated quite well through the mannequin.

felixrindt:
I also like outline. Something more subtle in any case.

PR #1608, opened 2025-07-27
fix acting user getting notifications

### LINKED ISSUE #1600: Responsibles get an email when they themselve dispatch another participant
**Describe the bug**

Responsibles get an email when they themselve dispatch another participant. So they receive a notification about an action they just performed themselves. I'd be interested to know if other people consider this unintended behavior.

**To Reproduce**
1. Go to a Disposition for an event you are responsible for.
2. Confirm a participant that is not you.
3. get a notification.

felixrindt:
we were comparing user model instance with ID :roll_eyes: 
closes #1600

### DIFF
--- a/ephios/core/signup/flow/base.py
+++ b/ephios/core/signup/flow/base.py
@@ -85,12 +85,14 @@ class BaseSignupFlow(AbstractSignupFlow):
             )
         return participation
 
-    def perform_decline(self, participant, participation=None, **kwargs):
+    def perform_decline(self, participant, participation=None, acting_user=None, **kwargs):
         """Create and configure a declining participation object for the given participant. `kwargs` may contain further instructions from a e.g. a form."""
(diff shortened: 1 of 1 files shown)


tests/core/test_notifications.py
```diff
@@ -22,161 +22,201 @@
...
+
+
+def test_responsibles_dont_get_notified_about_own_signup(django_app, event, planner):
+    django_app.get(
+        event.get_absolute_url(),
+        user=planner,
+    ).form.submit(name="signup_choice", value="sign_up").follow()
+    # even if responsible, the planner that signed themselves up should not be notified about their own action
+    call_command("send_notifications")
+    for m in mail.outbox:
```
felixrindt, line 161: These are the two new tests. The rest was moved out of the class that didn't do anything.

PR #1630, opened 2025-08-21
Qualification default expiration time

samuelson-ben:
Hi,
here a pr for the topic of the qualification requests.

As part of it, it would be suitable to have a default expiration time in the core so not every plugin as to build it itself.

See also https://github.com/ephios-dev/ephios-qualification-fixtures/pull/7 as the changes for the import repo.

Please look over it and give some feedback.

I know the way I solved the dilemma between having a fixed time and a set at the end of year isn't great but I don't want to introduce many new fields and it should be able to set something like the 15. May of the following year and end of year or month wouldn't be suitable.

I am open for other ideas.


jeriox:
Hi @samuelson-ben, we had another discussion about your proposal.

>     * If we introduce a default expiration to the model it should be part of the Qualification form in a way that non technical people can easily understand how it works. I don't think that's the case here. We could provide some multipart widget that offers selection of different modes.

We decided that we want to offer a multipart widget for the user. This can be a very simple version in the first iteration. We thought about radio buttons distinguishing between "after x years" and "at x.y. after z years". This should cover the common usecases and can still be extended later. The data should be stored in a JSONField in the database in the form of `{day: "1", month: "2", year: "+2"}`.

>     * model-wise, I'm not happy with having this complex datatype as a plain charfield with validator. Maybe we can encapsulate serialization, validation and functionality in a class with it's own Model and Form field. That would produce code like

 For convenience, this should be offered by a custom formfield which we can also reuse in other situations (e.g. #209).


Would you feel comfortable implementing that or should we support you in that regard?

samuelson-ben:
I cleaned up a bit, provided translations and added the new json format to the qualification-fixtures.
Currently I am looking for a good way to implement a function which takes a start date and returns and end date (date of the qualification -> expiration date). I haven't found a good way until now. Do you have ideas?

jeriox:
@samuelson-ben my PR is ready, feel free to check it out and merge it into your branch. Happy to hear your thoughts on it

jeriox:
Hi @samuelson-ben, just checking in if you want to take another look at my PR/the changes we proposed. If you prefer we could also finish your work. Either way would be cool to have your contribution included :)

PR #1644, opened 2025-09-05
underscore should alwasy be gettext_lazy

felixrindt:
If it's not lazy, strings specified in class attributes (like success_message on CBVs) are always translated in the default locale (de). `_` should always be the lazy variant to avoid confusion. In placed were it must not be lazy, use an explicit `gettext` instead. I used search to replace occurrences for the wrong import.

### DIFF
--- a/ephios/core/forms/users.py
+++ b/ephios/core/forms/users.py
@@ -15,7 +15,7 @@ from django.forms import (
     inlineformset_factory,
 )
 from django.urls import reverse
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext_lazy as _
 from django_select2.forms import Select2MultipleWidget, Select2Widget
 from guardian.shortcuts import assign_perm, get_objects_for_group, remove_perm
 
--- a/ephios/core/management/commands/devdata.py
+++ b/ephios/core/management/commands/devdata.py
@@ -4,7 +4,7 @@ from django.contrib.auth.hashers import make_password
 from django.core.management.base import BaseCommand
 from django.db import transaction
 from django.utils.timezone import make_aware
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext
 from guardian.shortcuts import assign_perm
 
 from ephios.core.models import Event, EventType, Shift, UserProfile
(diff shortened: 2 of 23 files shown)


lukasrad02:
Plain `gettext` is still used at <https://github.com/ephios-dev/ephios/blob/186c8e7ee227066cfa86bdd65850456d4f4c05ce/ephios/core/signup/views.py#L99>. I think that this is not an issue, as the message is evaluated within a view and thus uses the correct language, but is there any reason to use the non-lazy version here?

Another instance of plain `gettext` is used at <https://github.com/ephios-dev/ephios/blob/186c8e7ee227066cfa86bdd65850456d4f4c05ce/ephios/plugins/complexsignup/structure.py#L153>. As this is a class method, I think we'll have to use the lazy version here, but not familiar enough with the code to construct a scenario to trigger that code path.

lukasrad02:
And also some non-lazy `pgettext`/`ngettext`:

<https://github.com/ephios-dev/ephios/blob/186c8e7ee227066cfa86bdd65850456d4f4c05ce/ephios/core/models/events.py#L384>

<https://github.com/ephios-dev/ephios/blob/186c8e7ee227066cfa86bdd65850456d4f4c05ce/ephios/plugins/basesignupflows/flow/coupled.py#L60-L61>

<https://github.com/ephios-dev/ephios/blob/186c8e7ee227066cfa86bdd65850456d4f4c05ce/ephios/plugins/qualification_management/forms.py#L73>

felixrindt:
> As this is a class method, I think we'll have to use the lazy version here, but not familiar enough with the code to construct a scenario to trigger that code path.

As long as it's inside a request context, both is fine. It will be translated at some point. All locations you mentioned are evaluated inside a request context.

ephios/core/management/commands/devdata.py
```diff
@@ -23,15 +23,15 @@ def create_objects():
 
     from django.contrib.auth.models import Group
 
-    volunteers = Group.objects.create(name=_("Volunteers"))
+    volunteers = Group.objects.create(name=gettext("Volunteers"))
```
felixrindt, line 26: This was interesting: model char fields must get a str as value, not the proxy object. With the lazy version here, this later failed when `str(volunteers)` was called somewhere inside modellogging, because `__str__` did not return a `str` object, because `Group` simply returned the value of its name field.

ephios/core/forms/users.py
lukasrad02, line 1: In `get_group_permission_log_fields` in this file, neither `gettext` nor `gettext_lazy` are really fitting, since in both cases, the translation is fixed on the current locale of the acting user, instead of being determined when displaying the log entry:

Screenshot 2025-09-05 at 14-58-20 Home - ephios

(Feel free to move this into a separate issue if you don't find a quick way to fix this)

lukasrad02 (approved):
Lgtm

I found one occurrence where the translations are still not working correctly, but this is unrelated to the changes in this PR.

PR #1656, opened 2025-09-18
Use guardian PermissionRequiredMixin

felixrindt:


ephios/extra/mixins.py
```diff
@@ -1,39 +1,45 @@
...
-            return True
-        if not self.accept_object_perms or (obj := self.get_permission_object()) is None:
-            return False
-        return all(user.has_perm(perm, obj) for perm in perms)
+        if not self.accept_object_perms:
+            return None
+        return super().get_permission_object()
+
+
+class CustomCheckPermissionMixin(DjangoPermissionRequiredMixin):
```
jeriox: do we need this as a class? can't you just inherit from the django class and implement has_permission() where you are using it?

jeriox (approved):

PR #1718, opened 2025-12-18
View for sending custom mass emails

### LINKED ISSUE #1041: Streamline sending mails to a group of users
We have a couple places where we users can manually notify a group of users:

* announcing a new event
* event reminder
* mailing all participants of an event

And I think there should be another place to do that with the users filtered in the user list.

Now, I think it would be great in many cases if I could customize the message before sending and review the recipients in a separate view. All the buttons for the event notifications could still exist but would redirect to a view with subject, body and event prefilled. You could add/remove participants, customize the text and hit send.

felixrindt:
closes #1041 

* [x] issues für markdown preview field #1731
* [x] issue für email text tempaltes #1732
* [x] translation

### DIFF
--- a/ephios/core/services/notifications/types.py
+++ b/ephios/core/services/notifications/types.py
@@ -115,11 +115,13 @@ class FallbackNotification(AbstractNotificationHandler):
 
     @classmethod
     def get_subject(cls, notification):
-        return notification.data.get("subject")
+        return notification.data.get("subject", notification.slug)
 
     @classmethod
     def get_body(cls, notification):
-        return notification.data.get("body")
+        return notification.data.get(
+            "body", _("Unknown content for notification #{}").format(notification.pk)
+        )
 
     @classmethod
     def is_obsolete(cls, notification):
@@ -500,8 +502,8 @@ class SubjectBodyDataMixin:
 
 
 class GenericMassNotification(SubjectBodyDataMixin, AbstractNotificationHandler):
-    slug = "ephios_custom_event_reminder"
-    title = _("Mass mailing")
+    slug = "ephios_generic_mass"
+    title = _("You received a message from another user")
     unsubscribe_allowed = False
 
     @classmethod
@@ -533,7 +535,7 @@ class GenericMassNotification(SubjectBodyDataMixin, AbstractNotificationHandler)
 
 class CustomEventNotification(SubjectBodyDataMixin, AbstractNotificationHandler):
     slug = "ephios_custom_event_notification"
-    title = _("Information on an event")
+    title = _("A responsible shares information on an event")
     unsubscribe_allowed = False
 
     @classmethod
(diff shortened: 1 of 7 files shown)


ephios/core/services/notifications/types.py
```diff
@@ -107,6 +109,23 @@ def get_actions_with_referrer(cls, notification):
         return actions
 
 
+class FallbackNotification(AbstractNotificationHandler):
+    slug = "fallback"
+    title = "Fallback"
+
+    @classmethod
+    def get_subject(cls, notification):
+        return notification.data.get("subject")
```
jeriox: what about cases where there is no subject in the data?

ephios/core/services/notifications/types.py
```diff
@@ -514,101 +488,74 @@ def get_body(cls, notification):
...
+
+    @classmethod
+    def get_body(cls, notification):
+        return notification.data.get("body")
+
+
+class GenericMassNotification(SubjectBodyDataMixin, AbstractNotificationHandler):
+    slug = "ephios_custom_event_reminder"
```
jeriox: slug does not match title

ephios/core/templates/core/mass_notification_write.html
```diff
@@ -0,0 +1,153 @@
...
+    <script type="text/javascript" src="{% static "ephios/js/mass_notification_write.js" %}"></script>
+{% endblock %}
+
+{% block title %}
+    {% translate "Mass notification" %}
+{% endblock %}
+
+{% block html_head %}
+    {% if form.event %}
+        <style nonce="{{ request.csp_nonce }}">
```
jeriox: ugh i forgot that we had this ugly code. is there a way that we don't introduce this in another place?

ephios/core/templates/core/mass_notification_write.html
```diff
@@ -0,0 +1,153 @@
...
+
+{% block content %}
+    <div class="page-header">
+
+        <h1 class="card-title fw-bold fs-1">
+            {% translate "Mass Notification" %}
+        </h1>
+
+        <div class="alert alert-info">
+            {% if form.event %}
```
jeriox: remove

ephios/core/views/notifications.py
```diff
@@ -33,3 +50,182 @@ class NotificationMarkAllAsReadView(LoginRequiredMixin, RedirectView):
...
+            self.fields["to_participants"].help_text = _(
+                "You can only select users that have permission to view the event."
+            )
+
+    def _configure_choices(self):
+        choices = {}
+        self.participants_by_identifier = {}
+
+        # without event context, a user can send notifications to profiles they can see
+        user_qs = get_objects_for_user(user=self.request.user, perms=["core.view_userprofile"])
```
jeriox: maybe put this into else to prevent additional db query?

ephios/core/views/notifications.py
```diff
@@ -33,3 +50,182 @@ class NotificationMarkAllAsReadView(LoginRequiredMixin, RedirectView):
...
+            "subject": form.cleaned_data["subject"],
+            "body": form.cleaned_data["body"],
+        }
+        recipients = form.cleaned_data["to_participants"]
+        if self.event:
+            CustomEventNotification.send(self.event, recipients, **subject_and_body)
+            messages.success(
+                self.request,
+                ngettext_lazy(
+                    "Send notification to {count} participant.",
```
jeriox: senT?

PR #1730, opened 2026-01-10
prompt users for pwa installation and notifications

### LINKED ISSUE #1144: Improve UX for push notifications
taken from #989:

- I don't think users dig the settings to enable notifications. I think we should ask them when installing the PWA or after signing up for the first time or smth like that.
push notifications don't (always) use symbols or good deep links to be actionable... There are probably more features we aren't using.
- clicking the push notification could lead to a web version of the full notification to read it. That's probably better then directly redirecting to the disposition view if the users didn't read the whole notification or could not, because it was too long

jeriox:
closes #1144

### DIFF
--- a/ephios/templates/base.html
+++ b/ephios/templates/base.html
@@ -204,7 +204,7 @@
                     <p class="pb-2">{% blocktranslate trimmed with platform_name=platform_name %}
                         You can install {{ platform_name }} on your device and use it like any other app!
                     {% endblocktranslate %} {% translate "Simply click the button below to continue." %}</p>
-                    <button id="pwaInstall" class="btn btn-primary">{% blocktranslate with platform_name=platform_name %}Install {{ platform_name }}{% endblocktranslate %}</button>
+                    <button id="pwaInstall" class="btn btn-primary">{% blocktranslate trimmed with platform_name=platform_name %}Install {{ platform_name }}{% endblocktranslate %}</button>
                 </div>
             </div>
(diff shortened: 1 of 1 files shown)


ephios/static/ephios/js/main.js
```diff
@@ -155,8 +155,66 @@ $(document).ready(function () {
...
+    const notificationOffcanvas = document.getElementById("notificationOffcanvas");
+    const notificationBsOffcanvas = new bootstrap.Offcanvas('#notificationOffcanvas');
+    notificationOffcanvas.addEventListener("hide.bs.offcanvas", _ => rememberDismissed("notificationPrompt"))
+
+    let deferredInstallPrompt;
+
+    window.addEventListener('beforeinstallprompt', (e) => {
+      e.preventDefault();
+      deferredInstallPrompt = e;
+      if (!getCookie("pwaPrompt") && !window.location.pathname.startsWith("/accounts")) {
```
felixrindt: do you wanna exclude the login logout views or just sessions without login? There might be a better way, like a data value on <body>

ephios/static/ephios/js/main.js
```diff
@@ -155,8 +155,66 @@ $(document).ready(function () {
         });
     }
 
+    const pwaAndroidOffcanvas = document.getElementById('pwaAndroidOffcanvas');
```
felixrindt: I found your code here hard to navigate. Can you add comments to explain what each paragraph does and why? 
Also, where does isPushEnabled come from?

felixrindt (approved):

PR #1806, opened 2026-04-01
update pre-commit hooks, ruff to 0.15.8

felixrindt:


src/ephios/core/signup/flow/base.py
```diff
@@ -57,9 +56,10 @@ def configuration_form_class(self):
     def signup_action_validator_class(self):
         return BaseSignupActionValidator
 
-    @functools.lru_cache(1)
```
jeriox, line 60: is this out of fashion?

src/ephios/core/views/event.py
```diff
@@ -334,17 +334,24 @@ def _get_calendar_context(self):
...
         ctx = {}
         default_date = timezone.now()
         if self.request.session.has_key("day_calendar_last_date"):
-            default_date = datetime.strptime(
-                self.request.session.get("day_calendar_last_date"), "%Y-%m-%d"
-            ).date()
+            default_date = (
+                datetime
+                .strptime(self.request.session.get("day_calendar_last_date"), "%Y-%m-%d")
+                .replace(tzinfo=get_current_timezone())
```
jeriox, line 352: do we prefer this over django utils like make_aware?

jeriox (approved):

src/ephios/core/views/event.py
felixrindt: I think the linter doesn't understand the Django untils. I personally can better predict the results of the built in functions than the Django utils.

src/ephios/core/signup/flow/base.py
felixrindt: Can lead to memory issues: https://stackoverflow.com/questions/33672412/python-functools-lru-cache-with-instance-methods-release-object

PR #1808, opened 2026-04-03
pin pywebpush

jeriox:


pyproject.toml
```diff
@@ -76,6 +77,7 @@ dev = [
     "djhtml>=3.0.6,<4",
     "prek>=0.3.1",
     "ruff>=0.15.0",
+    "pre-commit>=4.5.1",
```
felixrindt: This is not a dependency, we changed this to prek.
```suggestion
```

felixrindt (approved):

PR #1809, opened 2026-04-03
add filter form to workinghour detail view, export workinghours

### LINKED ISSUE #838: Export working hours
As a resposible, I'd like the working hour overview to display an export button for the working hours. This should result in downloading a CSV or similar with the working hours to allow for further processing/statistics/diagrams. This could be implemented as an API route.

jeriox:
closes #1780 
closes #838


src/ephios/core/models/users.py
```diff
@@ -196,32 +197,40 @@ def qualifications(self):
             .select_related("category")
         )
 
-    def get_workhour_items(self):
+    def get_workhour_items(
+        self, start: date = date.min, end: date = date.max, eventtype: Optional[EventType] = None
```
felixrindt, line 201: Next to the red tests, with #1806 in mind, this will raise a linting error (use `T | None` instead of `Optional[T]`).

src/ephios/core/models/users.py
```diff
@@ -196,32 +197,40 @@ def qualifications(self):
...
+            participations = participations.filter(shift__event__type=eventtype)
+        participations = participations.annotate(
+            duration=ExpressionWrapper(
+                (F("end_time") - F("start_time")),
+                output_field=models.DurationField(),
+            ),
+            date=ExpressionWrapper(TruncDate(F("start_time")), output_field=DateField()),
+            reason=F("shift__event__title"),
+            type=Value("event"),
+            origin_id=F("shift__event__pk"),
```
felixrindt, line 221: any reason why this is after the if in l. 211 and not chained to the initial queryset definition?

src/ephios/core/views/workinghour.py
```diff
@@ -1,3 +1,5 @@
+import codecs
```
felixrindt, line 1: always wondered why this file is not `workingshours.py` (plural)

src/ephios/core/urls.py
```diff
@@ -269,6 +271,7 @@
         name="workinghours_request",
     ),
     path("workinghours/", WorkingHourOverview.as_view(), name="workinghours_list"),
+    path("workinghours/export", WorkingHourExportView.as_view(), name="workinghours_export"),
```
felixrindt, line 274:
```suggestion
    path("workinghours/export/", WorkingHourExportView.as_view(), name="workinghours_export"),
```

src/ephios/core/urls.py
```diff
@@ -278,6 +281,11 @@
         UserProfileWorkingHourView.as_view(),
         name="workinghours_detail",
     ),
+    path(
+        "workinghours/user/<int:pk>/export",
```
felixrindt, line 285:
```suggestion
        "workinghours/user/<int:pk>/export/",
```

src/ephios/core/views/workinghour.py
felixrindt, line 137: I think this view should also get the filter form. At my place we have this usecase where volunteers report hours in a specific event type for which they are eligible for compensation at the end of a year. That would be perfect for them to look up their own hours for this.

src/ephios/core/templates/core/userprofile_workinghours.html
```diff
@@ -12,14 +13,27 @@
...
-           href="{% url "core:workinghours_add" userprofile.pk %}">{% translate "Add working hours" %}</a>
+        <a class="btn btn-sm btn-secondary"
+           href="{% url "core:workinghours_add" userprofile.pk %}"><span class="fa fa-plus"></span> {% translate "Add working hours" %}</a>
+        <a class="btn btn-sm btn-secondary"
+           href="{% url "core:workinghours_detail_export" userprofile.pk %}"><span class="fa fa-download"></span> {% translate "Download list" %}</a>
     {% elif own_profile %}
-        <a class="btn btn-primary"
-           href="{% url "core:workinghours_request" %}">{% translate "Request working hours" %}</a>
+        <a class="btn btn-sm btn-secondary"
+           href="{% url "core:workinghours_request" %}"><span class="fa fa-plus"></span> {% translate "Request working hours" %}</a>
```
felixrindt, line 34: List export should also be available on your own profile imo.

src/ephios/core/views/workinghour.py
```diff
@@ -222,3 +243,64 @@ def _get_target_user(self):
...
+            eventtype=filter_form.cleaned_data.get("type"),
+        )[1]
+
+        rows = []
+        for entry in workinghours:
+            rows.append([
+                entry["date"],
+                entry["reason"],
+                timedelta_in_hours(entry["duration"]),
+                entry["type"],
```
felixrindt, line 293: I think it would be useful if - for events - this included the event type, so people could refine the analysis offline. 

Might even make sense to have it in the multi user export as columns. 🤔

felixrindt (changes requested):

src/ephios/core/views/workinghours.py
```diff
@@ -1,3 +1,5 @@
+import codecs
```
jeriox: we are not consistent at all. there is also event.py in the same folder

src/ephios/core/models/users.py
jeriox: the .value() has to be after the filter. but the annotation can move up, except for the fact that we then annotate a lot more rows than necessary

src/ephios/core/views/workinghours.py
felixrindt: well, feel free to change it at this point. (only if you feel like it)

src/ephios/core/models/users.py
felixrindt: I do mean the annotation. Order doesn't matter actually, we're just defining the queryset. The DBS optimizes the query when it is evaluated at `.values()`?!

felixrindt (changes requested):
Image

Image

Should opt for one version. Also, in my own participations (as a non-privilidged user) I don't have the download option, but I think I should!

src/ephios/core/models/users.py
```diff
@@ -194,32 +195,48 @@ def qualifications(self):
...
         hour_sum = (
             participations.aggregate(Sum("duration"))["duration__sum"] or datetime.timedelta()
-        ) + datetime.timedelta(
-            hours=float(workinghours.aggregate(Sum("duration"))["duration__sum"] or 0)
         )
+        workinghours = []
+        if eventtype is None:
+            workinghours = self.workinghours_set.annotate(
+                duration=F("hours"),
+                type=Value(_("Request"), output_field=CharField()),
```
felixrindt: I find `"Requests"` problemativ. Hours might have been manually added without the Consequence Request. Also, for translation, this should use p_gettext to provide context that this is not the same Request as in participation requests.

felixrindt (approved):
Besides that one Term lgtm

PR #1824, opened 2026-04-19
display email invalid warning to user

jeriox:


src/ephios/core/services/notifications/backends.py
```diff
@@ -102,7 +102,7 @@ def should_send(cls, notification):
     def user_prefers_sending(cls, notification):
         if not notification.user:
             return True
-        if not notification.user.is_active or notification.user.email_invalid:
```
felixrindt, line 105: So this is because we don't need to be careful anymore?

src/ephios/core/models/users.py
```diff
@@ -115,7 +115,7 @@ class UserProfile(guardian.mixins.GuardianUserMixin, PermissionsMixin, AbstractB
         choices=settings.LANGUAGES,
     )
     disabled_notifications = JSONField(
-        default=list, encoder=CustomJSONEncoder, decoder=CustomJSONDecoder
```
felixrindt, line 118: why does it need to be nullable? What does blank mean for JSONFields?
The default of an empty list should work, no?

src/ephios/core/forms/users.py
```diff
@@ -347,8 +347,7 @@ def save(self, commit=True):
                 Q(id__in=self.cleaned_data["groups"]) | Q(id__in=(g.id for g in self.locked_groups))
             )
         )
-        # if the user is re-activated after the email has been deemed invalid, reset the flag
-        if userprofile.is_active and userprofile.email_invalid:
+        if "email" in self.changed_data:
```
felixrindt, line 350: What if they cleaned up their inbox and we don't need to change the email?

felixrindt (approved):

Issue #620, opened 2021-11-26
Participation Consequences


felixrindt:
This is probably very low priority, but we could build consequences to alter participation state. Although we would probably want to display them in the disposition view etc.

* [ ] Consquences for declining when self-decline is not allowed.
* [ ] Consequences for individual times when providing individual times is not allowed
* [ ] https://github.com/ephios-dev/ephios/issues/66#issuecomment-763062635

PR #1826, opened 2026-04-20
allow users to customize confirmed participations

### LINKED ISSUE #1599: Allow participation customization also when users are not allowed to decline
Currently, a participation in a positive state can only be customized by the participant if they are also allowed to decline (https://github.com/ephios-dev/ephios/blob/937ec9b7b230c4ef8382d184f10a504382c9fea6/ephios/core/signup/flow/participant_validation.py#L202).

As a lot of new features are bound to the customize view (comments, questionnaires), we should also allow this in other cases. We need to make sure that the participant can only customize their times if this is allowed.

jeriox:
closes #1599

keeping the behaviour that users can customize times if they are confirmed and can decline and re-signup. we could also think about dropping this (related to #620)

### DIFF
--- a/src/ephios/core/signup/forms.py
+++ b/src/ephios/core/signup/forms.py
@@ -182,14 +182,19 @@ class SignupForm(BaseParticipationForm):
             FormActions(*self._get_buttons()),
         )
 
+        validator = self.instance and self.shift.signup_flow.get_validator(
+            self.instance.participant
+        )
+        can_customize_times = (
+            validator
+            and (not self.instance.is_in_positive_state() or validator.can_decline())
+            and not validator.get_signup_errors()
+        )
         if (
             not getattr(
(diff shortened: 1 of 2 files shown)


src/ephios/core/signup/forms.py
```diff
@@ -182,9 +182,19 @@ def __init__(self, *args, **kwargs):
             FormActions(*self._get_buttons()),
         )
```
jeriox, line 184: got a bit complex, we could also move that over to the validator e.g. as can_change_individual_times()

felixrindt: I would like to see that in the validator in case we introduce API-based signup, yes.

Alternatively, have you thought about replacing `can_customize_signup` as a `can_change_individual_time` and just always allowing people to open the customize form (if they can sign up or decline)?

PR #1841, opened 2026-05-17
add CacheControl Header and reload on bfcache restore

### LINKED ISSUE #1840: Browser back navigation creates desync local and server state after disposition edit
**Describe the bug**
After editing the disposition of a shift, the dispostion edit view is still in browser history. When navigating back to the event overview by using the back navigation twice, the local state, i.e., the changed disposition, is reverted back. After reloading the page server-side-state updates the local state accordingly

**To Reproduce**
1. Go to 'https://shifts.myhpi.de/events/181-hutmensch/' (or equivalent event overview)
2. Click on 'Disposition'
3. Change anything (for example decline a user)
4. Save the Changes.
5. Navigate back using browser back button twice (you should arrive back on the overview)
6. The disposition change is reverted in displayed UI**Expected behavior**
After editing a disposition, event or similar the edit-view gets removed from the browser navigation history.
7. Reload the page (should update back to server-side-state

**Expected Behavior**
The "Edit"-view should be removed from the browser navigation history.

**Environment**
Fedora Linux 44, Firefox

felixrindt:
closes #1840


src/ephios/extra/middleware.py
```diff
@@ -41,3 +42,15 @@ def __call__(self, request):
...
+
+class CacheControlMiddleware:
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response: HttpResponse = self.get_response(request)
+        # To prevent storing sensitive data in the cache as well as
+        # to prevent outdated data being shown to users
+        response.headers["Cache-Control"] = "no-cache, no-store, must-revalidate"
```
felixrindt, line 55: @jeriox says 10 mins TTL would also be fine. Does the browser even use the http cache for html documents?

